PR #7, opened 2015-04-30
rewrote plugin spec function AiDacrateChanged

cxd4:
This was all zilmar's original code for the function really; like every audio plugin uses it.  I just felt like it could be more readable to remove any potential distraction from things like, a bug in the main CPU emulator passing an invalid SystemType to the function, or oddball values of AI_DACRATE_REG.  I also felt like it may be more readable to invert the if { } else { } branch predictor so that the entire, much larger else { } bit doesn't all have to be indented an extra tab.  (See very first commit.)

Mostly a bunch of idealistic, minor changes, though I think it could help document some bits more.

So, for a few years now, I've always been curious what would happen on real hardware if a game managed in some way to access the reserved/unused bits of RCP registers.  For example, it's established that AI_DACRATE_REG has usually only the low 14 bits of merit [13..0].  I imagine that setting anything else shouldn't have an effect, but, just in case of potentially unexplored possibilities like that, I felt like adding some verbosity there.

### DIFF
--- a/AziAudio/main.cpp
+++ b/AziAudio/main.cpp
@@ -299,6 +299,15 @@ EXPORT void CALL AiDacrateChanged (int  SystemType) {
 		return;
 
 	Dacrate = *AudioInfo.AI_DACRATE_REG & 0x00003FFF;
+#ifdef _DEBUG
+	if (Dacrate != *AudioInfo.AI_DACRATE_REG)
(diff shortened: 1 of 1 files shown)


Azimer:
Looks good to me.  The plugin could use a little more robustness.

PR #9, opened 2015-04-30
got the plugin to compile in MinGW GCC/G++

cxd4:
I wanted to give it a try to see how hard it might be to get this plugin to compile using MinGW.  I was afraid it would be a wasted effort, but actually it just took a few changes (less than 20 generally small and divided commits, outlined below).

I guess I'm mostly a portability geek, so it was partly for fun, but then again, there are both pros and cons to linking Windows binaries with MinGW instead of Microsoft Visual C++.  (50% of the time I'd say the pros outweigh the cons 100% of the time.)  At the very least, it opens up the possibility for testing the plugin built by a different compiler, which might help isolate whether some bugs are problems in the C++ code or unspecified behavior that varies between compiler to compiler.

Unfortunately, I do not have `XAudio2.h` handy on me in my compiler toolchain...I did however have access to `dsound.h` and libdsound.a, so I was able to compile and link this plugin using MinGW for DirectSound.  I wish that in the future the same can be said for XAudio2...a quick search shows that some people have made it happen, but I have yet to play with that.

Needless to say it took more changes than are usually done in one straightforward pull request, so I'll try to resolve any potential issues in any of my commits or answer any questions that may need answering.

### DIFF
--- a/AziAudio/ABI1.cpp
+++ b/AziAudio/ABI1.cpp
@@ -247,9 +247,6 @@ void ENVMIXER () {
 				mov eax, dword ptr [LRamp];
 				imul ecx;
 				shrd eax, edx, 16;
-				//shl edx, 16;
-				//shr eax, 16;
-				//add eax, edx;
 				mov dword ptr [LAdderEnd], eax;
 				mov eax, dword ptr [LAdderStart];
 				mov dword ptr [LAcc], eax;
@@ -272,9 +269,7 @@ void ENVMIXER () {
 				mov ecx, dword ptr [RAdderEnd];
 				mov eax, dword ptr [RRamp];
 				imul ecx;
-				shl edx, 16;
-				shr eax, 16;
-				add eax, edx;
+				shrd eax, edx, 16;
 				mov dword ptr [RAdderEnd], eax;
 				mov eax, dword ptr [RAdderStart];
 				mov dword ptr [RAcc], eax;
--- a/AziAudio/ABI1.cpp
+++ b/AziAudio/ABI1.cpp
@@ -531,7 +531,7 @@ void RESAMPLE () {
 	}
 
 	if ((Flags & 0x2))
-		__asm int 3;
+		x86_interrupt();
 
 	for(int i=0;i < ((AudioCount+0xf)&0xFFF0)/2;i++)	{
 		//location = (((Accum * 0x40) >> 0x10) * 8);
--- a/AziAudio/ABI1.cpp
+++ b/AziAudio/ABI1.cpp
@@ -952,7 +952,7 @@ void LOADADPCM () { // Loads an ADPCM table - Works 100% Now 03-13-01
 	v0 = (t9 & 0xffffff);// + SEGMENTS[(t9>>24)&0xf];
 #ifdef _DEBUG
 	if (v0 > (1024*1024*8))
-		v0 = (t9 & 0xffffff);*/
+		v0 = (t9 & 0xffffff);
 	memcpy (dmem+0x4c0, rdram+v0, k0&0xffff); // Could prolly get away with not putting this in dmem
 	assert ((k0&0xffff) <= 0x80);
 #endif
--- a/AziAudio/ABI1.cpp
+++ b/AziAudio/ABI1.cpp
@@ -547,8 +547,7 @@ void RESAMPLE () {
 			src[(srcPtr+x)^1] = 0;//*(u16 *)(rdram+((addy+x)^2));
 	}
 
-	if ((Flags & 0x2))
-		x86_interrupt();
+	assert((Flags & 0x2) == 0);
 
 	for(int i=0;i < ((AudioCount+0xf)&0xFFF0)/2;i++)	{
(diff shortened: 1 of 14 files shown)


Azimer:
I'm guessing it was an error MSVC let me do.  I think I meant something like...
```
void SETBUFF () { // Should work ;-)
	if ((k0 >> 0x10) & 0x8) { // A_AUX - Auxillary Sound Buffer Settings
		AudioAuxA		= (u16)(k0);
		AudioAuxC		= (u16)((t9 >> 0x10));
		AudioAuxE		= (u16)(t9);
	} else {		// A_MAIN - Main Sound Buffer Settings
		AudioInBuffer   = u16(k0); // 0x00
		AudioOutBuffer	= (u16)((t9 >> 0x10)); // 0x02
		AudioCount		= (u16)(t9); // 0x04
	}
}
```
As far as the in-line assembly code.  All that code can go away.  We shouldn't need it since machines are sufficiently fast enough for N64 emulation.

I'll take a closer look at the changes later but I did want to respond to your comments right away.

cxd4:
>I'm guessing it was an error MSVC let me do.

Did recall seeing it a bit in Project64 Glide64 source as well ... u16(exp) instead of (u16)exp.

So I think I will append one more commit to change all the u16(exp) to (u16)(exp).

>As far as the in-line assembly code. All that code can go away. We shouldn't need it since machines are sufficiently fast enough for N64 emulation.

+ one more as well to flatten the asm/no-asm to just the C version I wrote

AziAudio/ABI3mp3.cpp
```diff
@@ -282,7 +296,7 @@ void InnerLoop () {
 				RSP_Vect[0].DW[0] = 0x0002D4130005A827;
 */
 				if ((t1 | t2 | t3 | t5 | t6) & 0x1)
-					__asm int 3;
+					x86_interrupt();
```
Azimer: Change to assert(true);

AziAudio/ABI3mp3.cpp
```diff
@@ -236,7 +236,21 @@ void MP3 () {
 	}
 }
 
-
+void x86_interrupt(void)
```
Azimer: Remove no longer needed.

AziAudio/ABI3.cpp
```diff
@@ -515,8 +512,10 @@ void SAVEBUFF3 () {
 void LOADADPCM3 () { // Loads an ADPCM table - Works 100% Now 03-13-01
 	u32 v0;
 	v0 = (t9 & 0xffffff);
-	//memcpy (dmem+0x3f0, rdram+v0, k0&0xffff); 
-	//assert ((k0&0xffff) <= 0x80);
+#ifdef _DEBUG
+	memcpy (dmem+0x3f0, rdram+v0, k0&0xffff);
+	assert ((k0&0xffff) <= 0x80);
+#endif
```
Azimer: Comment this out... even for a _DEBUG build.

AziAudio/ABI1.cpp
```diff
@@ -943,10 +950,12 @@ void DMEMMOVE () { // Doesn't sound just right?... will fix when HLE is ready -
...
-/*	if (v0 > (1024*1024*8))
-		v0 = (t9 & 0xffffff);*/
-	//memcpy (dmem+0x4c0, rdram+v0, k0&0xffff); // Could prolly get away with not putting this in dmem
-	//assert ((k0&0xffff) <= 0x80);
+#ifdef _DEBUG
+	if (v0 > (1024*1024*8))
+		v0 = (t9 & 0xffffff);
+	memcpy (dmem+0x4c0, rdram+v0, k0&0xffff); // Could prolly get away with not putting this in dmem
+	assert ((k0&0xffff) <= 0x80);
+#endif
```
Azimer: Leave these commented out even for a _DEBUG build.  The issue here is we do not need to put this in dmem and the extraneous assert may cause errors when the code is only checking to see if the adpcm table is at least 0x80 in length.  I suppose it's possible other games may not use a table this long.

AziAudio/ABI1.cpp
```diff
@@ -928,8 +933,10 @@ void DMEMMOVE () { // Doesn't sound just right?... will fix when HLE is ready -
 		return;
 	v0 = (k0 & 0xFFFF);
 	v1 = (t9 >> 0x10);
-	//assert ((v1 & 0x3) == 0);
-	//assert ((v0 & 0x3) == 0);
+#ifdef _DEBUG
+	assert((v1 & 0x3) == 0);
+	assert((v0 & 0x3) == 0);
+#endif
```
Azimer: The assert on line 937 is causing Mario64 to be asserted.  We need to suppress these.

Azimer:
I will approve once these changes are completed.  Thanks for your contribution!

cxd4:
I think the last two commits I just pushed (above this post I am making, below your last one) seem to have taken care of the rest of that stuff you said.

Azimer:
Thanks for your extra work.

PR #12, opened 2015-05-02
I don't always clamp, but when i do...

cxd4:
...I like to have a central [macro/inline] function for doing it with SSE. :)

The performance may not necessarily be any better though, because right now all I did was install my `pack_signed` function throughout ABI1, ABI2, and ABI3.  Loading a 32-bit word to the low 128 bits of an XMM, doing a SIMD signed pack of the whole thing, just to store only one 32-bit result seems a bit costly.  However, it definitely is at least more static/fixed than saying `if (x < -32768) x = -32768; else if (x > 32767) x = 32767;` all the time.  The problem is that SSE instructions use more byte code...AziAudio.dll grew from 65.0 KB to 66.0 KB as a result of installing my function.

It goes back down again to 65.0 though if you enable the macro versions of my functions (instead of saying inline functions) though, which goes back to the original dynamic if/else algorithm of doing it.  So at least it's maintainable then.

Another solution (to get the DLL size back down, *plus* extra perf.) would be to start using the vector pack function I also installed:  `vsats128`, which finally does make use of the SSE2 large byte code size by doing 4 packs at once.  However the ABI code might become slightly less readable (or maybe more), because then you can't say o1/a1/a2/a3 all the time; you have to move everything to arrays.

### DIFF
--- a/AziAudio/ABI1.cpp
+++ b/AziAudio/ABI1.cpp
@@ -345,14 +345,10 @@ void ENVMIXER () {
 		AuxL  = (((s64)Wet*2 * (s64)(LAcc>>16)) + 0x8000) >> 16; 
 		AuxR  = (((s64)Wet*2 * (s64)(RAcc>>16)) + 0x8000) >> 16;*/
 /*
-		if (MainL>32767) MainL = 32767;
-		else if (MainL<-32768) MainL = -32768;
-		if (MainR>32767) MainR = 32767;
-		else if (MainR<-32768) MainR = -32768;
-		if (AuxL>32767) AuxL = 32767;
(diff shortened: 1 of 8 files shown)


Azimer:
I will merge this after I get a chance to do some regression testing later today.

cxd4:
As per your feedback here:  https://github.com/Azimer/AziAudio/issues/13#issuecomment-98488197
I will do one more commit to apply my clamping function also over the API3mp3 clamp to -32767.

cxd4:
^ Pushed the commit.

Theoretically shouldn't be any regressions (except of course that last commit I just now did, though hopefully that's a fix instead of a regression), but needless to say in this line of work we can never be too careful.

PR #14, opened 2015-05-02
MinGW build fixes for some old versions of Windows

cxd4:
Frank74 pointed out a few things I overlooked while making the build script.  I have not used XP in a while so wasn't aware until now.

The only unresolved thing is the GUI (and still no XAudio2)...I can't figure out why it's not working.  I did at least include the resource script in the linker config this time though, which I forgot to do last time.  I use the same config on my angrylion RDP plugin to get the GUI to work, but why it doesn't work here, I'm not sure.  May be different rules when linking C++ objects than C objects or something.

### DIFF
--- a/make_w32.cmd
+++ b/make_w32.cmd
@@ -1,9 +1,9 @@
 @ECHO OFF
 TITLE MinGW Compiler Suite Invocation
 
-set MinGW=C:/MinGW
-set src=%CD%/AziAudio
-set obj=%CD%/gccbuild
+set MinGW=C:\MinGW
+set src=%CD%\AziAudio
+set obj=%CD%\gccbuild
 
 if not exist gccbuild (
 mkdir gccbuild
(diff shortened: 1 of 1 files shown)


Frank-74:
It builds fine now.  But why is my AziAudio.dll 566Kb?  I see you mention 65-66Kb in the clamp PR.
Still no config though.

cxd4:
That is quite surprising.  I have two theories behind how that can be:
1.  g++ on your end isn't invoking `strip` from the `-s` flag I passed to the linker, which can strip hundreds of KB of debugging information.
2.  The C++ runtime libraries are being statically linked.  MinGW linker does things a little bit differently than the Microsoft Visual C++ MSVCP*, MSVCR* libraries...if you compile as C++ then a few extra linker plugins get involved, some unusual DLL files but ones we would rather link dynamically than statically.

For that reason just got to work on one more commit before any of this should get merged.

cxd4:
@Frank-74 any idea what version of MinGW you're using?

http://cboard.cprogramming.com/cplusplus-programming/97009-cplusplus-runtime-dynamic-link-mingw.html
http://www.ciiycode.com/0SmNmqeUgUqe/mingw-471-doesnt-recognize-command-line-option-staticlibstdc.html

It seems that in some older versions of MinGW, you're not allowed to choose.  It forces statically linking them for you, whereas for my MinGW 4.8.2 installation it's always been dynamically linking it by default.  If I specify `-static-libstdc++` to force linking it statically, it comes up to larger than a 600 KB DLL, similar to what you reported.

Try that last commit I just pushed now; I don't know if it will work on your version of the linker.

cxd4:
I've just added one more commit to decrease the DLL size even further.

For me, it drops the MinGW build size from 90 KB to 70 KB exactly.
The trick was that I was using -O3 to optimize the entire plugin, which is often over-aggressive and risky with its optimizations.  I changed the ABI HLE stuff to -O2, which is similar but safer (still better than MSVC's compiler optimizations though :P) and the non-RSP-relevant stuff (trivial things like main.cpp, context management) to -Os which optimizes for small binary size.

Frank-74:
I had 4.7.1, just uninstalled completely and re-installed the latest repository version 4.8.1-4.  Now AziAudio.dll is only 72,192 bytes.  Strange that...

cxd4:
Alright, well I did what I could to prevent statically linking that library.  I was just hoping it couldn't cause the build to fail on older versions like 4.7.1 like you were using.  However upgrading MinGW is at least a good a solution then again.

So that just leaves the plugin configuration dialog UI...which for the life of me I cannot understand what is wrong.  I always got resource UIs linking fine with a couple other plugins I did so why not this one?  @LegendOfDragoon are you able to determine what might be the missing link with getting the plugin config showing up if you run the batch file for the MinGW build?

Azimer:
Is this PR ready to merge?  I hate to say I know nothing of MinGW. :)

cxd4:
Well.  All I can say is that the only things missing are XAudio2 and, mysteriously, the config UI.
This pull request was meant mostly to fix the script to the way Windows XP used to interpret things, and I've succeeded in that department.  I think it should be ready to merge.

I've tried linking in the resource file the same way I used Visual Studio to do it in my other projects, but for some confounding reason the audio plugin configuration form isn't popping up on the MinGW build.  I really know nothing of Windows API stuff when it comes to resources and things like that, so I am not sure that I'm the best person to debug what is failing there.

XAudio2 on the other hand I'm fairly sure there is something to be done about that so that will work, but that will take time and I have to upload libxaudio.a to the repository so MinGW will be able to find that.  I was thinking that investigation could be separate from this pull request and fixed in a different one.

As for MinGW, it has its problems.  It's what they get for taking the standard Unix compilers and doing a minimalist Linux abstraction of it to a native Windows port of the compiler--however it does have its advantages, like better control over the optimization than what MSVC does.

cxd4:
@Frank-74 : We've found out that our missing config issue with audio plugin was caused actually from outdated Win32 headers before updated MinGW packages.

If you do a manual installation of the individual Win32 headers from here:
http://sourceforge.net/projects/mingw/files/MinGW/Base/w32api/
http://sourceforge.net/projects/mingw/files/MinGW/Base/w32api/w32api-4.0.3/
http://sourceforge.net/projects/mingw/files/MinGW/Base/w32api/w32api-4.0.3/w32api-4.0.3-1-mingw32-src.tar.lzma/download

Replace all of your MinGW/include files with those of the include files available from that archive.
Then recompile and you should have a config UI now?

Azimer:
@cxd4 : Can you describe what is going on with MinGW and XAudio2?  I wonder if it's something similar where it's trying to use the API version of XAudio2 2.8 instead of the COM version of XAudio2 2.7.  The former will only work under Windows 8 and requires a library xaudio2.lib.  Missing the library gives you an error similar to https://github.com/Azimer/AziAudio/issues/19.

cxd4:
The issue currently is that there are compile-time errors--it seems to compile well enough with the MinGW-standard `xaudio.h` include I bundled with this repo, but that requires some other header file (forgot exactly...can look it up later but something related to COM declarations) which was spewing out loads of compile-time errors.

Past that, I just need to find the semi-unofficial MinGW-compatible (MinGW authors don't supply an official one for us) libxaudio.a library to link against to get this plugin working with XAudio2.  I am meaning to try that when I wake up.

Azimer:
Sounds good.  So completely unrelated. :smile_cat:

Frank-74:
@cxd4 Yes, I have a config UI now.  Thanks.

Also, you've included an older version of xaudio2.h from August 2009 DX SDK.  Maybe add the path $(DXSDK_DIR)\Include.  That's the path used in codeblocks for the DX SDK.

cxd4:
Yeah that's probable.

I'm not familiar with the different versions of XAudio2 API.  I grabbed the header from a package that was using MinGW to compile XAudio2 for a separate project, but I still have to make sure it stops spewing out all those COM declaration errors (if we remove the -DXAUDIO_LIBRARIES_UNAVAILABLE line from this batch script) and am not using wrong version of the header or something.  So I'll check out the Code::Blocks one and see if that helps.

>Sounds good. So completely unrelated. :smile_cat:

dat cat tho.

Issue #3, opened 2015-04-29
Bug associated with Initiating Audio without a sound card present


Azimer:
Implement silent playback emulation to allow the audio emulation to continue to working

PR #15, opened 2015-05-02
Prevent all the XAudio2 and DirectSound crashing when playing with no sound.

cxd4:
If you have no sound card or you're like me and accidentally yank your headphones unplugged every once in a while, you may have noticed that this plugin (either in DirectSound or XAudio2 mode) will crash at various points as early as when starting a ROM and as late as when shutting down the emulation thread.  This pull request is meant mostly to just address that.

It doesn't fix this issue still:
https://github.com/Azimer/AziAudio/issues/3

I know because I boot up Rogue Squadron, but it's hung up waiting for an interrupt from the audio interface.  This happens when sound is playing, but not when playing with my headphones unplugged.

### DIFF
--- a/AziAudio/DirectSoundDriver.cpp
+++ b/AziAudio/DirectSoundDriver.cpp
@@ -385,7 +385,10 @@ BOOL DirectSoundDriver::Initialize(HWND hwnd) {
 
 	WaitForSingleObject(hMutex, INFINITE);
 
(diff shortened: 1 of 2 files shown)


Azimer:
Seems okay.  Thanks.  
I hope we can nail your nosound issue soon and close https://github.com/Azimer/AziAudio/issues/3

Issue #2, opened 2015-04-29
Word Driver Championship and Stuntracer


Azimer:
These games sound awful.  An on-demand option needs to be implemented, one that is compatible with other audio playback.

cxd4:
Hm.  Has anyone else also been able to get either of these games booting with the latest AziAudio, or is it just me?  Need some more time but so far PJ64 just crashes with "Unknown memory action"; maybe it's the latest project64.exe I'm using or something.

Azimer:
I'm pretty sure it worked with Project64 2.1.  I've never gotten an unknown memory action though so that is new to me if this is the case.  I should have a moment to sit down in the next couple days to retest unless someone else gets to it first.

LegendOfDragoon:
For me it does not work in 2.1. XA2 WIP2 still works on PJ64 2.2, so I don't think it has anything to do with the emulator. The DS8 version works as well.

I first noticed this issue in XA2 WIP 4. As a reminder, i mentioned it here http://www.emutalk.net/threads/55161-Audio-v0-70-WIP-4-Feedback/page8

Azimer:
Weird.  I will compile from the source and give it a try.  When I tested, I tested a release from last August.

cxd4:
I can't very well remember @LegendOfDragoon, but weren't some of you commenting on an issue with the Project64 repository where RomOpen was having an issue because of something to do with something in Azimer's plugins?  Somebody ended up changing RomOpen with the core's handling of the audio spec at some point weeks ago, but I can't remember exactly when or what commit.  Might have to do some bunches of `git checkout` to narrow when this began exactly.

LegendOfDragoon:
That was for loading save states. For some reason, XA2 WIP4 (and 5) crashes when loading save states on any emulator that uses RomClose during the loadstate process. I believe the deinitialization / initialization code needs to be looked at, in Azimer's XA2 code.

cxd4:
It does boot if I turn on HLE gfx with Jabo's 1.7 plugin though.
Seems it's just that if I use LLE gfx plugins with WDC/SR64 then it crashes as I said.

LegendOfDragoon:
>It does boot if I turn on HLE gfx with Jabo's 1.7 plugin though.
Seems it's just that if I use LLE gfx plugins with WDC/SR64 then it crashes as I said.

That's weird. Doesn't work for me at all.

Azimer should be able to easily fix this problem though. He just needs to fix initialization.

[...]

PR #16, opened 2015-05-03
Fix initialization problem in WDC with XA2

LegendOfDragoon:
The crash occurred on line #230, when booting up WDC.


Azimer:
This fixes the symptom but not the problem.  I know what's wrong but this is a good fix until that primary issue is fixed.  The issue regarding WDC https://github.com/Azimer/AziAudio/issues/2 will stay open.  I will leave a comment on the issue so it's not lost.

PR #17, opened 2015-05-03
Add Sleep(1) to AudioThreadProc in DS8

LegendOfDragoon:
Without Sleep(1), a whole cpu core is used up, without any noticeable
benefit, afaik.


Azimer:
This is fine and an oversight.  While testing I sometimes change it to Sleep(0) to know the audio thread is not causing the delay/issues.

PR #18, opened 2015-05-04
Replaced block of code in AB1, AB2, AB3 with for loop

samuelyuan:
The code fits a pattern, so a for loop would be much more concise. Please look it over. I'm sure that the for loop does the exact same thing, just with way less code.


Azimer:
Thanks for the PR.  I will do regression testing prior to committing.

cxd4:
Appears to me to still be correct, nice idea.

I think a function name would go even better, since this basic audio HLE pattern is used between the different microcodes.  So I think taking this a step further and creating a function may be even better--not just remove more lines of code but document with a good name of the function exactly what phase of the HLE this basic pattern represents for the 3 ABI's?

samuelyuan:
I'm not exactly sure what this part of the code is supposed to do.

Azimer:
I am closing this PR.  I cherry-picked the first change but I won't accept the other.  The trouble I have is the redundancy and it's not going in the direction I wish to take the code.  I am thinking we'll want to separate out the commands from ABI1-ABI3.cpp into seperate command files envmixer.cpp, mixer.cpp, adpcm.cpp so it's easier to take out what's in common.  I will put up an issue to define this work.  Thanks @samuelyuan for your contribution!

Azimer:
I added your ADPCMFillArray function to my internal build I hope to put in a PR for later this evening.  Thanks again for your contribution.  In hindsight, I may have been better off merging the entire PR then moving it later.  Sorry for the confusion.

PR #33, opened 2015-05-05
Slightly less stdint dependency

cxd4:
Well actually, what I did was remove *some* stdint dependency (the INT16_MIN and _MAX macros), and replace the reference behind the rest with my own header, `my_types.h`.  The basic, primary difference between `my_types.h` and `mytypes.h` is that the former detects if `<stdint.h>` integer limits were defined to determine whether or not the header has been included yet, and if it hasn't, then include `<stdint.h>` *if we can prove that it most likely must exist*, or else define my u16, i32, i64 etc. types into the uint16_t etc. types ourselves.  It's more advanced than the `mytypes.h` header Azimer's been using, so I was thinking maybe we should replace it?

It was also meant to be a central pillar header between the 4 plugin type headers (gfx.h, audio.h, contr.h, rsp.h) from my rcp64 repo, though I don't really care if it's named `my_types.h` or `mytypes.h` as long as it's consistent in both repositories.

According to @Frank-74 the other thing that MSVC 2008 will have trouble with is stdbool.h or something?  But I am not as experienced with that one as I never use `bool`, and it looks like that header was included locally by bobby smiles already.  If there are any questions or suggestions let me know; I'm sure there are a few things I probably could have done different.


Azimer:
Thanks.  Will merge without testing but reading compares everything passes my sanity check. :+1:

cxd4:
Whoops.  Mysteriously I now get missing "mytypes.h" errors after pulling from your merge.

It was testing fine when I made the PR, but I think what happened was I waited too long to pull your latest file split updates to my master branch, so one more PR is in order to adjust to this.  Sorry if this threw anyone off.

PR #36, opened 2015-05-05
XAudio2SoundDriver: Fix duplicate conditions

lioncash:
I'm assuming this was a typo.


Azimer:
I think this was an oops. :+1:

PR #37, opened 2015-05-06
Vectorize some nasty LWC2/transpose and MAC stuff in ABI_Filters.cpp.

cxd4:
Well, here's one that I spent a few more hours on.  Now that the rest of the source has been addressed enough to my liking, I decided to start cleaning up some RSP stuff.  I was most interested in `ABI_Filters.cpp` because I thought it made use of the vector shuffle operands, but it turned out to be 16-bit mixed-/bi-endianness within 32-bit word vector tricks in LWC2 pack operations.  (Man those suck.)  In the end I was able to make a function out of most of that.

I'm not sure yet which games use ABI2 ucode, though, but I'll try to figure that out in the meantime to verify the consequences of my changes.  Ever since the very first commit, I already started to think a little regression testing wouldn't hurt.  I'll address any bugs in these commits if they exist.

### DIFF
--- a/AziAudio/ABI_Filters.cpp
+++ b/AziAudio/ABI_Filters.cpp
@@ -49,84 +49,84 @@ void FILTER2() {
 	inp2 = (short *)(BufferSpace + inPtr);
 	for (x = 0; x < cnt; x += 0x10) {
 		out1[0] =
-			inp2[1] * lutt6[0] +
 			inp2[0] * lutt6[1] +
-			inp1[7] * lutt6[2] +
+			inp2[1] * lutt6[0] +
 			inp1[6] * lutt6[3] +
-			inp1[5] * lutt6[4] +
+			inp1[7] * lutt6[2] +
 			inp1[4] * lutt6[5] +
-			inp1[3] * lutt6[6] +
-			inp1[2] * lutt6[7]
+			inp1[5] * lutt6[4] +
+			inp1[2] * lutt6[7] +
+			inp1[3] * lutt6[6]
 		;
 		out1[1] =
-			inp1[6] * lutt6[0] +
 			inp2[1] * lutt6[1] +
-			inp1[4] * lutt6[2] +
+			inp1[6] * lutt6[0] +
 			inp1[7] * lutt6[3] +
-			inp1[2] * lutt6[4] +
+			inp1[4] * lutt6[2] +
 			inp1[5] * lutt6[5] +
-			inp1[0] * lutt6[6] +
-			inp1[3] * lutt6[7]
+			inp1[2] * lutt6[4] +
+			inp1[3] * lutt6[7] +
+			inp1[0] * lutt6[6]
 		;
 		out1[2] =
-			inp2[3] * lutt6[0] +
 			inp2[2] * lutt6[1] +
-			inp2[1] * lutt6[2] +
+			inp2[3] * lutt6[0] +
 			inp2[0] * lutt6[3] +
-			inp1[7] * lutt6[4] +
+			inp2[1] * lutt6[2] +
 			inp1[6] * lutt6[5] +
-			inp1[5] * lutt6[6] +
-			inp1[4] * lutt6[7]
+			inp1[7] * lutt6[4] +
+			inp1[4] * lutt6[7] +
+			inp1[5] * lutt6[6]
 		;
 		out1[3] =
-			inp2[0] * lutt6[0] +
 			inp2[3] * lutt6[1] +
-			inp1[6] * lutt6[2] +
+			inp2[0] * lutt6[0] +
 			inp2[1] * lutt6[3] +
-			inp1[4] * lutt6[4] +
+			inp1[6] * lutt6[2] +
 			inp1[7] * lutt6[5] +
-			inp1[2] * lutt6[6] +
-			inp1[5] * lutt6[7]
+			inp1[4] * lutt6[4] +
+			inp1[5] * lutt6[7] +
+			inp1[2] * lutt6[6]
 		;
 		out1[4] =
-			inp2[5] * lutt6[0] +
 			inp2[4] * lutt6[1] +
-			inp2[3] * lutt6[2] +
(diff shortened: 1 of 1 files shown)


Azimer:
I agree so don't mind me taking my time to test. ABI2 tend to be games like mario kart, zelda and other first party titles.

cxd4:
Simpler titles than I thought ... I'll check those out and keep that in mind.

Assuming I didn't glitch out during my matrix inversions in all the previous commits, I'm guessing that if there is a bug, it probably is in my last commit, where I removed your (s16) type converter--I took a gamble that since your comment implied that clamping was the intention, it shouldn't be needed since true clamping emulation should require more than that (and never happens in any commercial games using VMULF anyway that I've seen so far).

Azimer:
If I recall correctly, it was more a note for myself to let me know these values need to be clamped.

cxd4:
Hardware operation says they need to be clamped, but if this is based on VMULF or VMADN or anything like that, I can say from experience that games never seemed to exercise out-of-bounds values that warrant the requirement to emulate this clamping, for those particular ucodes.  So I'm hypothesizing that you can probably omit clamping here entirely, but if I'm wrong, change my #if 0 to an #if 1 in the last commit.

LegendOfDragoon:
Maybe we can have debug builds that people can test out, to see if some of these optimizations are safe to make :smile:. Even though I'm all for optimizing, I try to debug and see if it changes the results of the equation. If it's only a very slight deviation, like the current algorithm for Mixer, then I keep the optimization enabled, generally.

Azimer:
I guess I am just showing my age and 20+ years of looking at pointers as *. :)

Anyway a more pressing matter.  Zelda OOT intro has a regression in this PR.  I can't accept it in its current form.  You can hear it in the very beginning part of the intro with the horse gallop and music.

cxd4:
>I guess I am just showing my age and 20+ years of looking at pointers as *. :)

It really is mostly an effect of better communication of ideas, syntax and grammar based on habit, experience, and practice.  I actually have an old habit of always writing `return;` at the end of all functions, even if they are `void` functions, simply because it is an old habit of mine and looks more readable to me, personally, than it probably does to you.  (After all, every function must have a return point, or execution leaks past the end of the function into the unknown if you don't do a `return;`, or the compiler should generate one for you.)

I don't have near 20+ years of experience like you do, although I do have a lot of experience as a beginner trying desperately to understand pointer syntax in C.  I myself never understood saying `int* p` or `int * p` and only ever considered `int *p;` ... now I am the exact opposite and can only understand the first two but not the third.  It's a consequence of finally realizing how complicated pointers in C **aren't**, or at least *shouldn't be*.  Why do beginners have to learn how to use `*` both for pointer reference AND declaration?  Why can't we just declare memory addresses as `unsigned int address` and reference as `*(address)`?  The linguistic answer (technical issues with address type sizes set aside) is, I believe there is no reason why we can't, and the asterisk is an arbitrary choice of operator to people trying to understand when it's actually required for reading from memory to actually work.

Anyway I'm fine with both methods, and my `pi16` type is harder for people who are used to seeing the `*` to know it's a pointer and inconsistent with the rest of your source, so I happily changed that back.

cxd4:
Also the more pressing matter that you mentioned was taken care of.

I didn't notice you were using pointer variables this extensively near the end of the entire loop, hence the regression--I usually never increment pointer addresses and just the index for an array to refer to them, but your method is better performance.

Azimer:
Sounds good.  You didn't need to change the pi16 back.  I am not such an old fossil I couldn't adjust. :)  Zelda OOT sounds good again with this change.  Thanks!

cxd4:
Yes I believed you when you said you weren't insistent on `pi16` warranting a reverse change, but I realized that it matched the rest of the practice in the plugin anyway and often say `i16 * p;` myself sometimes in my own projects (as long as it's not two pointers declared in one line).

AmbientMalice:
@cxd4 I'm not sure exactly where or how, but this PR seems to have broken save state loading. Games are mute after loading save states.

cxd4:
I have no problems with sound after loading saved states in Project64 1.6.
I can't get anything at all with states to load in pj64 2.x so can't verify what you're saying until I pull from the latest changes to get the latest EXE.  (And no, a buildbot would not help with the issue, cause I'd still have to download an EXE file every single day then in that case.)

In the meantime talk about what games exactly you mean?  Because the contents of this pull request affect only games using ABI2 ucode afaik ... so if it affects Mario64 or something else then you're probably looking at some other commit to blame there.

AmbientMalice:
@cxd4

You're right. I did some more testing. The commit at fault is https://github.com/Azimer/AziAudio/pull/40 -- the problem is XAudio, basically. Before this commit, I get a Dsound plugin. Afterwards, the only plugin at the end of build is the XAudio one.

The XAudio version will crash on save state load if FAT isn't enabled. If FAT is enabled, games will be mute.

cxd4:
I forgot you use MinGW like me to test sometimes.

That pull request you linked to of mine was to get XAudio2 headers supplied and working with MinGW...I suppose a few of what I changed could affect MSVC 2013 build as well.  Do you have save state issue with 2013 build

cxd4:
>Before this commit, I get a Dsound plugin. Afterwards, the only plugin at the end of build is the XAudio one.

If you were to add -DXAUDIO_LIBRARIES_UNAVAILABLE^ to somewhere underneath my FLAGS_x86 script variable you could still test MinGW build with DirectSound if you wanted.

AmbientMalice:
@cxd4
>If you were to add -DXAUDIO_LIBRARIES_UNAVAILABLE^ to somewhere underneath my FLAGS_x86 script variable you could still test MinGW build with DirectSound if you wanted.

DS version works fine. Now that I think about it, surely the problem is that XAudio Azimer's has been broken in PJ64 2.x for a while now?

https://github.com/project64/project64/issues/115

As with related issues, resetting with F1 before loading save states will prevent the crash. (Horrible crackling sound after loading save states, though. At least when I tested.)

PR #39, opened 2015-05-06
Separate Dewindow function

samuelyuan:
The InnerLoop in ABI3mp3 is extremely bloated as of now. First, I separated a block of code into a separate function. Then the actual multiplication is separated into another function.

### DIFF
--- a/AziAudio/ABI3mp3.cpp
+++ b/AziAudio/ABI3mp3.cpp
@@ -192,6 +192,121 @@ void MP3AB0 () {
 	}
 }
 
+void Dewindow()
+{
+	//Part 8 
+
+	//u64 *DW = (u64 *)&DeWindowLUT[0x10-(t4>>1)];
+	u32 offset = 0x10 - (t4 >> 1);
+
+	u32 addptr = t6 & 0xFFE0;
+	offset = 0x10 - (t4 >> 1);
+
+	s32 v2 = 0, v4 = 0, v6 = 0, v8 = 0;
+	s32 z2 = 0, z4 = 0, z6 = 0, z8 = 0;
+
+	offset = 0x10 - (t4 >> 1);// + x*0x40;
+	for (int x = 0; x < 8; x++) {
+		v2 = v4 = v6 = v8 = 0;
+
+		//addptr = t1;
+
+		for (int i = 7; i >= 0; i--) {
+			v2 += ((int)*(s16 *)(mp3data + (addptr)+0x00) * (short)DeWindowLUT[offset + 0x00] + 0x4000) >> 0xF;
+			v4 += ((int)*(s16 *)(mp3data + (addptr)+0x10) * (short)DeWindowLUT[offset + 0x08] + 0x4000) >> 0xF;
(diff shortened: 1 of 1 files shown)


Azimer:
I have some constructive criticism and perhaps @cxd4 can weigh in here based on his optimization experiences.  But if you pack things into functions for the sake of putting it into functions, what are you really gaining other than adding using extra CPU cycles on overhead?  I'd like to have a constructive debate on this because I am not convinced it's more understandable.

samuelyuan:
Fair enough, my understanding was that functions should not be excessively long. However, the way I did it might not be the best. It might be more readable to some, but make no difference to others. So if there is a better way, then I would like to know.

LegendOfDragoon:
I'm not a fan of making functions that you only call once. Although occasionally I'll still do it. I personally like to minimize cpu usage, but it seems that most people do not care about it as much as I do :) .

However, that CalcDeWindow seems ok to me. Just not sure about the Dewindow function.

Azimer:
I do care about performance and I am hoping we can add some SIMD support after we get things cleaned up a little.  Otherwise I am fine with the way it was because I understand the code perfectly since I reversed engineered and wrote it :laughing:

cxd4:
Generally, if there is a function that you call only once, then any modern optimizing compiler should (in my experience) be less likely to *not* inline the function.  Because, one of the chief benefits you gain from NOT inlining functions, is repeated re-usability of the same instruction cache space.  That advantage no longer applies if you declare a function and call it only once.

There is still a chance that GCC etc. will still refuse to inline it--like if you took 5000 lines of code out of an 8000-line function or something, then it would keep the 5000 lines to the function call, and I think that is the correct decision for cacheability.  When you rationalize how the code is contacted at that point--by procedural call or static in-lining--against the entirety of all the work to be done, performance so far as that particular step goes becomes an inferior measure.

It's part of why I used a function pointer table for my RSP vector instructions emulator but a switch() jump table for the scalar, regular MIPS R4000 ops emulation--plain MIPS ops are much more direct to emulate and use less instruction cache space in the interpreter function to cycle-between, whereas vector ops emulation uses a lot of byte code and for cacheability I kept those to function tables for speed.

But yeah, I don't really like the name of the `Dewindow` function...I try to avoid creating too many functions from this code without knowing exactly what's going on.  I don't like to pick generic names until I've finished figuring out what is being accessed from the smaller individual steps.  So I think it makes more sense to allocate functions for the smaller, highly repeated steps first, and save any large functions like `Dewindow` for last after understanding the bits and pieces.  In that sense I think the second commit of this PR looks good--but should have come long before the first commit and sort of has a backwards order look to it.

samuelyuan:
Based off of the suggestions here, I reverted just the first commit. I kept the other function CalcDeWindow though.

Azimer:
Okay.  I can't test this change but it looks okay.  I will merge with the revert.  Thanks!

Azimer:
Also I'd like to add that it's okay to have dialog in this PRs prior to committing your changes.  You can easily put in a PR, discuss your changes, commit then discuss what needs to be done before accepting the change.  The difference being, it may save you from having to do reverts and muddling up the history.  No big deal this time but just saying it's option.  It could also be open as an issue prior to any code being laid down too.

PR #41, opened 2015-05-07
Share DirectX headers between MinGW/g++ and VS2008.

cxd4:
What do we think about having a new folder named "3rd Party" in the root of the repository for having the DirectX headers in them?  In exchange, I can get rid of the `gccbuild` directory and files from this repo now since its only existence is as a result of running my make script anyway.

I was thinking it could be helpful to everyone if some synchronization was attempted to our guest fork over at project64:  https://github.com/project64/AziAudio

Here zilmar has the DirectX headers instead under a folder, `3rd Party`.  I figure it probably is misleading that I had them under a folder named `mingw_dx` since really they should be usable outside of just MinGW and work as convenience headers for people who prefer to compile with MSVC 2008 but just didn't happen to already have the DirectX SDK installed.  So this way we could share the headers between compilers, if somebody decided to add it to the include path in their own VS solution.

What do we think, should it be `3rd Party/directx/include`, just `directx/include` or something else?  I chose "3rd Party" just to match the way it's currently been done but we could change it.  I figured it could be misleading since they're not really "third-party" to this plugin as it is an XAudio plugin...however as @project64 pointed out while we were asking about the synchronicity it does emit the correct impression of keeping people interested away from those headers since they should not normally need changes like the rest of the actual source to the plugin.


Azimer:
Sounds good to me and makes perfect sense.  Now that I am home, I am going to work on verifying your other PR before it gets out of sync and requires a manual merge before anything else gets merged.

cxd4:
heh.  well it's a good thing I changed only the ABI filters.cpp file in that other PR...so hopefully it's unlikely we'll see any commits to create a merge conflict there, therefore should be easy to take our time making sure I didn't overlook anything.

PR #46, opened 2015-05-07
Have frequency-rounding performed on leveled intervals, through a function.

cxd4:
Not too sure about this one for a couple reasons, but I wanted to see how appropriate it could be to just have a function to address the code complication issue mentioned here?  https://github.com/Frank-74/AziAudio/commit/300d8523c1b51c2b38f5efde762d2f96e75b63bf#commitcomment-11086246

There may be a static, branch-free approach at this too, but atm it's just a function for portability.

Aside from the risk of maintenance with explicit type-cast conversions, my uncertainty with these commits lies in this field here from the comments that I replaced:
```c
} else if (Frequency < 15000 && Frequency > 10000) {
	Frequency = 11025;
```
Here, `Frequency` could be 14999 (almost 4000 Hz more than the original 11025-Hz target approximation) and get saturated to 11025 as a result of this block, but if it was 7025 (4000 Hz **less** than the original) there is no way it is going to be brought closer to the valid 11025 value.  So that is the sort of effect that my function replaces...but, I figured that my method could hopefully be more accurate due to the balanced approach with just a single range deviation rather than an arbitrarily specified min/max like 10000/15000?

### DIFF
--- a/AziAudio/DirectSoundDriver.cpp
+++ b/AziAudio/DirectSoundDriver.cpp
@@ -459,26 +459,28 @@ void DirectSoundDriver::DeInitialize() {
 
 // Buffer Functions for the Audio Code
 void DirectSoundDriver::SetFrequency(DWORD Frequency2) {
+	long freq;
 	DWORD Frequency = Frequency2;
 	BOOL bAudioPlaying = audioIsPlaying;
+
 	printf("SetFrequency()\n");
 	// MusyX - (Frequency / 80) * 4
 	// 
 	if (bAudioPlaying == TRUE) {
 		StopAudio();
 	}
-	/*
-	if (Frequency < 45000 && Frequency > 43000) {
-	Frequency = 44100;
-	}
-	else if (Frequency < 33000 && Frequency > 31000)	{
-	Frequency = 32000;
-	}
-	else if (Frequency < 25000 && Frequency > 20000)	{
-	Frequency = 22050;
-	} else if (Frequency < 15000 && Frequency > 10000) {
-	Frequency = 11025;
-	}*/
+
+	assert(Frequency <= 65535);
+	freq = (long)Frequency;
+
+	freq = filter_range(freq, 44100, 1000);
+	freq = filter_range(freq, 32000, 1000);
+	freq = filter_range(freq, 22050, 1000);
+	freq = filter_range(freq, 11025, 1000);
+
+	Frequency = (DWORD)freq;
+	assert(Frequency <= 44100);
+
 	sLOCK_SIZE = (Frequency / 80) * 4;// 0x600;// (22050 / 30) * 4;// 0x4000;// (Frequency / 60) * 4;
 	SampleRate = Frequency;
 	SegmentSize = 0; // Trash it... we need to redo the Frequency anyway...
--- a/AziAudio/SoundDriver.h
+++ b/AziAudio/SoundDriver.h
@@ -16,6 +16,11 @@
 #define SND_IS_NOT_EMPTY 0x4000000
 #define SND_IS_FULL		 0x8000000
 
+/*
+ * Filters an integer for proximity within the range (base +/- leeway).
+ * Created mostly just to solve the deviant N64 frequencies problem.
+ */
+extern long filter_range(long x, long base, int leeway);
 
 class SoundDriver
 {
--- a/AziAudio/XAudio2SoundDriver.cpp
+++ b/AziAudio/XAudio2SoundDriver.cpp
@@ -188,7 +188,7 @@ void XAudio2SoundDriver::SetFrequency(DWORD Frequency)
 	Frequency = (DWORD)freq;
 	assert(Frequency <= 44100);
 
-	cacheSize = (Frequency / 25) * 4;// (((Frequency * 4) / 100) & ~0x3) * 8;
+	cacheSize = (freq / 25) * 4;// (((Frequency * 4) / 100) & ~0x3) * 8;
 	if (Setup() < 0) /* failed to apply a sound device */
 		return;
 	g_source->SetSourceSampleRate(Frequency);
--- a/AziAudio/XAudio2SoundDriver.cpp
+++ b/AziAudio/XAudio2SoundDriver.cpp
@@ -173,21 +173,22 @@ void XAudio2SoundDriver::Teardown()
 
 void XAudio2SoundDriver::SetFrequency(DWORD Frequency)
 {
-//	Frequency = (Frequency+25) - ((Frequency+25) % 50);
-	if (Frequency < 45000 && Frequency > 43000) {
-		Frequency = 44100;
-	}
-	else if (Frequency < 33000 && Frequency > 31000)	{
-		Frequency = 32000;
-	}
-	else if (Frequency < 25000 && Frequency > 20000)	{
-		Frequency = 22050;
-	}
-	else if (Frequency < 15000 && Frequency > 10000) {
-		Frequency = 11025;
-	}
-	cacheSize = (Frequency / 25) * 4;// (((Frequency * 4) / 100) & ~0x3) * 8;
+	long freq;
 
+//	Frequency = (Frequency+25) - ((Frequency+25) % 50);
+
+	assert(Frequency <= 65535);
+	freq = (long)Frequency;
+
+	freq = filter_range(freq, 44100, 1000);
+	freq = filter_range(freq, 32000, 1000);
+	freq = filter_range(freq, 22050, 1000);
+	freq = filter_range(freq, 11025, 1000);
+
+	Frequency = (DWORD)freq;
+	assert(Frequency <= 44100);
+
+	cacheSize = (Frequency / 25) * 4;// (((Frequency * 4) / 100) & ~0x3) * 8;
 	if (Setup() < 0) /* failed to apply a sound device */
 		return;
 	g_source->SetSourceSampleRate(Frequency);
--- a/AziAudio/main.cpp
+++ b/AziAudio/main.cpp
@@ -465,3 +465,15 @@ int safe_strcpy(char* dst, size_t limit, const char* src)
     return (failure);
 #endif
 }
+
+long filter_range(long x, long base, int leeway)
+{
+    const long minimum = base - leeway; /* does not prevent signed underflow */
+    const long maximum = base + leeway; /* does not prevent signed overflow */
+
+    if (x < minimum)
+        return (x); /* filter_range(44095, 44100, 4) should return 44095. */
+    if (x > maximum)
+        return (x); /* filter_range(44105, 44100, 4) should return 44105. */
+    return (base);
+}


Azimer:
We can give it a shot.  From my experience (and the reason I put in the quick and easy code) was most games are wither 11025, 22050, 32000 or 44100Hz.  I know if they submit 32000, majority of the time it's set as 32006, 22050 will be 22048, etc.  Your way is definitely better for that instance where they may decide they want to do something a bit off.

cxd4:
Yeah that's the way I remember it.  With Jabo's wav logger it was always 32006 Hz, 44095 Hz ...

Then I tested with LaC's audio wav logger for Nemu64 and it was like, 44106 Hz, 3199? something or whatever?  So I dunno if there's a different way to calculate the frequencies but yeah.

LegendOfDragoon:
Dezaemon 3D has a sample rate of 15997, and Duck Dodgers has a sample rate of 24004. So I think you should add those in.

cxd4:
They were not supported previously, so I don't see hard evidence that they should be supported now.

I mostly replicated the frequency support Azimer had already in the comments.  Technically I'm not even sure that any of these frequencies are more hardware-accurate after rounding them from 32006 to 32000 etc., but I'm sure round frequencies are nice to have for the API's smoothness especially after what Frank74 went through.  Maybe he can try some stuff using other sample rates like you mentioned and if so we can add those in later.

Frank-74:
I've been experimenting for hours with this.  Finally I have the best result, which stops the clicking sounds in V-Rally Edition 99 (U).
Simply remove this,
```
+ freq = filter_range(freq, 44100, 1000);
+ freq = filter_range(freq, 32000, 1000);
+ freq = filter_range(freq, 22050, 1000);
+ freq = filter_range(freq, 11025, 1000);
```
And replace with just this.
```
	freq = filter_range(freq, 44100, 0);
```

theboy181:
You figure out Top Gear Rally out yet?

Frank-74:
@theboy181 No, still sounds bad.

cxd4:
>And replace with this.
```v
    freq = filter_range(freq, 44100, 0);
```

That means
```c
if (freq == 44100)
    freq = 44100;
else
    freq = freq;
```
which is a null operation.  Which means if you remove that line as well, there is no change.

LegendOfDragoon:
>Technically I'm not even sure that any of these frequencies are more hardware-accurate after rounding them from 32006 to 32000 etc.

Yesterday I went and checked the frequencies for a bunch of games. I'm starting to think it might be best not to even round frequencies or only round specific ones like 11023, 22047, 32006, and 44095. I'm pretty sure there's a reason a game decided to do weird ones like 21617. I initially thought rounding to 50 was good, but didn't know that games like Dark Rift used 11023.

> but I'm sure round frequencies are nice to have for the API's smoothness especially after what Frank74 went through.

Considering Goldeneye's frequency never changes, during the part he recorded (the intro), changing the rounding code made absolutely no difference because the result was the same, regardless (22050).

cxd4:
If it helps @Frank-74 the `filter_range` function I wrote is simple:  It takes parameter 1 and compares it to parameter 2, within a +/- range parameter 3.

filter_range(freq, 44100, 0) means see if freq is within +/- 0 (third parameter) off of 44100, so if freq = 44100 it is, if freq != 44100 it isn't and freq = freq.  I created this function for rounding; if you say 0 for the third parameter there is absolutely no change.  So what you said suggests that you experience the best audio quality from no frequency rounding at all?

Frank-74:
@cxd4 Yes, its best not to do any frequency rounding at all.  I've removed the new functions and just kept this for SetFrequency.  Its the best quality.
```
void XAudio2SoundDriver::SetFrequency(DWORD Frequency)
{
	cacheSize = (Frequency / 25) * 4;//(((Frequency * 4) / 100) & ~0x3) * 8;

	if (Setup() < 0) /* failed to apply a sound device */
		return;

	g_source->SetSourceSampleRate(Frequency);
}
```
The instance any frequency change happens, it resamples to the frequency rounded to and back to 44100 all the time.  Which causes degradation in sample quality.  Also causes clicks when the new frequency is set.

Frank-74:
I'm going to try and change when `SetFrequency` is called, because the docs say, 'to use SetSourceSampleRate the voice must not have any buffers currently queued'.  I have a feeling the clicking without FAT=0 is because of this.

cxd4:
Maybe the functions can be removed then, but they were committed in to replace the if-else style of doing it that Azimer had in comments, which you un-commented as part of your audio fix PR.  Therefore it would seem preferable to comment out my functions instead of delete, or better yet #if 0 macro it out.

>The instance any frequency change happens, it resamples to the frequency rounded to and back to 44100 all the time.

The un-rounded frequency should always from my records have been 44095 Hz (unless you use LaC's nemu audio), so I guess from what you are saying, games sound better with 44095 Hz instead of 44100 Hz.

LegendOfDragoon:
I think @cxd4 's code should stay in, just maybe change 1000 to a lower number.

cxd4:
If I change 1000 to a lower number, it will no longer be like the commented-out code that Azimer had before Frank un-commented it in his PR.

What Frank74 is saying is that the audio sounds best when you leave the frequencies un-rounded--my function calls exist to round the frequencies; if he says they sound best un-rounded then my new function is useless in all ways but except to rewrite the old code Azimer commented out (which I think he said he commented because it was a bunch of if-else not anywhere static like a switch statement for example).  So assuming this is all very well and true (32006 Hz sounds better than 32000 Hz etc.), then I suggest either putting my replacement functions in comments as well like Azimer had it, or doing an #if 0 around it so people can toggle the 0 quickly for testing.

Frank-74:
I've just quickly made a branch with the frequency rounding removed for testing.  I just commented it out.

LegendOfDragoon:
@cxd4 , It was mostly fine before tbh ;/ . These recent changes are what caused issues in VRally with FAT + AI = 0. So 1000 is indeed too high.

Frank-74:
@LegendOfDragoon If a sample of 44095 or 32006 is rounded to 44100 and 32000, then the sample is at the wrong rate for playback, which can make it sound grainy and lose attack.  And when the change happens, it introduces clicks.

Anyway, can you test SSB and tell me if Mario and Luigi's voices sound like chipmunks?  I've been told they do, but they sound the same pitch as Jabo's here.  Cheers.

theboy181:
Sorry, the chipmunk thing was PJ related. I reset the emulator, and I was getting the same performance as the comment I pointed out..

Do you have any examples of games that are different on the changes you made? something as clear to hear as V-Rally?

I like the idea off exact frequency's, why do we want to round them in the first place? If rounding is a must why wouldn't we round to the closest 25?

Frank-74:
Ok, I've created a PR to just change the leeway from 1000 to 10.  And added all the rates supported by my soundcard below 44100.  Just in case some game might use them.
https://github.com/Azimer/AziAudio/pull/55

LegendOfDragoon:
@Frank-74 I appreciate it :smile: .

Here's some data I collected the other day btw, for frequencies. I ran a bunch of games and looked for unique Sample rates.

```
DAC   Sample Rate   Games
4415     11023      Dark Rift
3042     15997      Dezaemon 3D
2535     19196      Lode Runner
2433     20000      Indy Racing
2251     21617      VRally
2212     21998      WDC & Top Gear
2210     22018      Conker, Mickey
2207     22047      Rogue Squadron and many others
2204     22077      Zool - Majuu Tsukai Densetsu
2163     22496      Polaris SnoCross
2027     24004      Duck Dodgers
1815     26807      Mario Kart
1689     28805      WCW NWO Revenge, WrestleMania, No Mercy, Pokemon Stadium 2
1520     32006      Super Smash Brothers and many others
1139     42703      International SuperStar Soccer
1103     44095      Mystical Ninja, Aidyn Chronicles, + some others
```

Frank-74:
What role does the DAC number play?

I can see its related, high sample rates have lower DAC numbers.  Is it the time/clock for playback at the sample rate?  AIDacrate?  If it is, rounding the sample rate could mess with timing of the audio.  Some samples might not finish before the DAC expires.  Its probably what causes the clicks when the leeway is changed, on occasions when the audio has already finished before the DAC has.

Just guessing.  I'd prefer not to round the sample rate at all if this is so.

LegendOfDragoon:
@Frank-74 idk why I wrote DAC xD. Changed it to Dacrate. I included that value because I've seen an audio plugin use 
```
Frequency = 48681812 / (Dacrate + 0);
instead of 
Frequency = 48681812 / (Dacrate + 1);
```

Frank-74:
@LegendOfDragoon Then would it be better to set the sample rate from the Dacrate numbers instead?
```
    Frequency = video_clock / Dacrate;
    g_source->SetSourceSampleRate(Frequency);
```
Thats what it does on AiDacrateChanged.  Or change the `Frequency = video_clock / (Dacrate + 1);` in AiDacrateChanged to match the rounded frequency.

I don't know anymore... :)

cxd4:
>@cxd4 , It was mostly fine before tbh ;/ .

Why tell me this?  I didn't change whether it was commented out or not, just the method of writing it.

cxd4:
>@LegendOfDragoon If a sample of 44095 or 32006 is rounded to 44100 and 32000, then the sample is at the wrong rate for playback, which can make it sound grainy and lose attack. And when the change happens, it introduces clicks.

Why change the leeway from 1000 to 10 then?  If rounding the output frequency throughout the game from 32006 to 32000 is bad I would think just 0 or no function call at all (better yet commented out) would be correct.

LegendOfDragoon:
>Why tell me this? You suggest this as if I created a regression.
I didn't change whether it was commented out or not, just the method of writing the same algorithm.

I know you didn't create a regression. I was just pointing out that having a high number for the leeway causes problems with Fixed Audio Timing in games like VRally. Not sure exactly what to do with rounding, because it's hard for me to tell a difference when you change it very slightly.

cxd4:
Having a high number for the leeway has no modified effect from having a low number for it.

If the third parameter is 1000, 44095 is still within 1000 of 44100.
If the third parameter is 10, 44095 is still within 10 or 44100.
You need to set it to 4 or less or remove the function completely for it to be a different result.

LegendOfDragoon:
The problem is that VRally has a Sample Rate of 21617, since it's within the 1000 range, it gets set to 22050 and that causes problems with FAT + AI = 0.

Frank-74:
@LegendOfDragoon 
How did you collect that list of Dacrate/Sample Rates?
I'm wanting to make a list for PAL versions.  To compare the frequency calculation.  As some games have much better sync/timing in PAL.

Cheers.

LegendOfDragoon:
@Frank-74 , I simply found the memory address of the variables and used cheat engine, so that I can see values in real time. After adding the addresses to a cheat engine table, I was able to quickly boot up each game and see the Dacrate and Sample Rate. The easiest way to find the address is to set a breakpoint around the code where that variable is. If you want, I can make the cheat engine table for you, just let me know what exact version of the source to compile.

I set the update interval in cheat engine to 1ms, so that it's more responsive. I love cheat engine :smile: .

Frank-74:
Well I put your table in a spreadsheet and started messing with the video clock.
```
Dacrate	NTSC VC	Freq 	NTSC Dac+1	Games
4415	48686400	11025		Dark Rift
3042	48686400	15999.4742	Dezaemon 3D
2535	48686400	19198.10726	Lode Runner
2433	48686400	20002.62942	Indy Racing
2251	48686400	21619.18295	VRally
2212	48686400	22000.18075	WDC & Top Gear
2210	48686400	22020.08141	Conker, Mickey
2207	48686400	22050		Rogue Squadron and many others
2204	48686400	22080		Zool - Majuu Tsukai Densetsu
2163	48686400	22498.33641	Polaris SnoCross
2027	48686400	24007.10059	Duck Dodgers
1815	48686400	26809.69163	Mario Kart
1689	48686400	28808.52071	WCW NWO Revenge, WrestleMania, No Mercy, Pokemon Stadium 2
1520	48686400	32009.46746	Super Smash Brothers and many others
1139	48686400	42707.36842	International SuperStar Soccer
1103	48686400	44100		Mystical Ninja, Aidyn Chronicles + some others
```

PR #47, opened 2015-05-08
Got it compiling on VS2008.

cxd4:
Both the DirectSound8 and XAudio 2.7 builds of the VS2008 Express solution I've set up on this pull request seem to work fine for _Gauntlet Legends_ and _Super Mario 64_, that I tested.

I was probably supposed to merge the differences in @project64's version of the VS2008 solution file he got working but just haven't gotten to that yet.  I actually replaced my .sln with his and my .vcproj with his and it spurred out a bunch of thousands of winapi header errors for some reason, plus I don't know if there's any advantage to making the solution with VS Express instead of VS Professional like he used, plus a few other differences I noticed (PropertySheets?  what's that).  These can be addressed over time.

EDIT:  Only the Debug build of the DirectSound plugin plays sound, not the Release build.  The Release build segfaults from reading a NULL pointer from some DirectSound call if a game tries to play audio.  Have yet to debug that issue.


Azimer:
Can you ensure master compiles for you still after the XBox commit from weiner?

cxd4:
Yes I will check this shortly...

cxd4:
It does compile still on VS2008, but I'm still going to chance a few things in an upcoming PR.

cxd4:
http://gameservice.googlecode.com/svn/trunk/Utils.h

According to this, if `_XBOX` is defined (targeting Xbox), we can still use the Microsoft built-in LLP64 API types __int8, __int16, __int32 and __int64 instead of any custom stdint.h header anyway.  Please correct me sooner than later @weinerschnitzel if I'm mistaken.

PR #48, opened 2015-05-08
Store values in mp3data using a function

samuelyuan:


samuelyuan:
Oops, I only meant to add one file.

PR #49, opened 2015-05-08
Store values in mp3data using a function 

samuelyuan:


samuelyuan:
I'm not sure what just happened, but something's not right.

cxd4:
That looks like a sticky situation.  I'm not a Git expert but it looks like you may have pulled all upstream changes to your master branch, then pushed to your master branch of the fork before doing any future pull requests.  In practice I find it to be a good idea to use separate branches and not just my master branch.

If in doubt/unfamiliar with Git your only option may be to delete your fork of AziAudio, re-fork AziAudio to your GitHub, and then clone that to your repo to start over then do the PR.

PR #50, opened 2015-05-08
Preliminary XBox port changes

weinerschnitzel:
Here are the first of my changes to get the audio plugin working with Surreal64. The changes should be neutral and would ultimately benefit me from constant rebasing.


Azimer:
I think I'm okay adding an xbox solution.  However, I do want you to be aware that you will be the one to support it and keep it up to date with code changes.  I also want everyone else to be aware of these defines and changes so they aren't removed out of confusion.  I am just not so sure about an XBox directory under the main source because it has nothing to do with the main source.  Thoughts?

weinerschnitzel:
I would like to keep a stubs file so I don't end up having to indef a lot of the source code. Although I am only using it for MessageBox atm. I can take another look at the stdint.h dependency and try to get away with just HatCat's my_types.h, but the question arises for where to put the XBox dependencies. I personally like a platform specific folder in the same location as the project file, but maybe we can start a AziAudio/Make folder and move the project files and XBox depps from AziAudio/AziAudio?

I plan to keep the XBox port up-to-date, the DirectSound changes should be small enough to not scratch heads.

Azimer:
I am going to merge this as is.  We can move stuff around later if needed.  If we do anything that requires your involvement to fix the solution/project (like moving that XBox folder), we can either tag you in the PR or attempt to fix it ourselves prior to submitting it.  @cxd4  @LegendOfDragoon  @samuelyuan - Tagged as an FYI since you're the most active contributors.

Thanks!

Azimer:
I had conflicts but I resolved them the best I could.  Please review.

PR #51, opened 2015-05-08
Add a function to save a value into mp3data

samuelyuan:


samuelyuan:
For some reason, the previous two pull requests didn't merge correctly with the master branch. Please disregard those pr's. This one should be the correct one.  

Sorry for all that confusion :(

Azimer:
No problem.

PR #52, opened 2015-05-08
Balance the portability changes with some Xbox-/Win32-agnosticism restored.

cxd4:
I have some kind of pseudo-migraine headache left over so don't feel like going into a lot of depth on this one at the moment, but basically I wanted to clean up a few of the recent Xbox-ifications to not be so strongly emphasized throughout the source.  (Granted it's hardly any less evil to do the same thing with Windows API code and use of the `windows.h` header, but that's an assumed fault since naturally this plugin always uses DirectX for audio.)

It also adjusts what @weinerschnitzel's pull request likely *would* have been if it had been made based on my last pull request, which I can tell that it wasn't (aside from the merge conflict Azimer had with merging), so that is updated as well.  Most importantly, please avoid imposing `FUNC_TYPE()` and `NAME_DEFINE()` in the plugin spec definitions--zilmar never did do a very good job of flexibility in his spec headers, but as far as I can see nothing warrants the need for introducing these macro functions.  I took them back out and updated my `my_types.h` header to give the old `EXPORT` and `CALL` macros null definitions when compiling on Xbox etc. to prevent the errors that way instead.  Your thoughts on any of this please weinerschnitzel?  Do you have any suggestions for or against my changes?


weinerschnitzel:
Well, NAME_DEFINE() and FUNC_TYPE() are to give each plugin a unique set of function names when linking statically for XBox. It makes it easier to link a binary with several audio plugins rather than build one binary for each plugin combination. 

I need the unique names. I suppose those functions could be wrapped and a separate XboxAudioSpec.h could offer the unique names if you think the NAME_DEFINE() and FUNC_TYPE() should go

cxd4:
Why do we need FUNC_TYPE() then?  Why does that exist?
```c
#ifdef _XBOX
#define NAME_DEFINE(name) _AUDIO_AZIAUD_##name
#define FUNC_TYPE(type) type
#else
#define NAME_DEFINE(name) CALL name
#define FUNC_TYPE(type) EXPORT type
#endif
```
All that does is remove the EXPORT #ifndef _XBOX, and EXPORT is the plugin spec macro for maintaining how the function symbol should be available.  Wouldn't it make more sense to just have an empty macro definition for EXPORT than to create a new function macro just to remove it?

weinerschnitzel:
You're right. FUNC_TYPE() isn't necessary. If you are ok with leaving NAME_DEFINE() and null defining EXPORT and CALL, that works for me too.

cxd4:
Okay, got it.  Sounds like the NAME_DEFINE thing is because from the sound of things, Surreal has plugin options statically linked?  Because if they are dynamically linked then I'm not sure why function names need to all be unique between plugins.

If the issue is static linking then I might be forced to put FUNC_TYPE() in, I'm thinking about it.

cxd4:
@weinerschnitzel : I added one more commit just now.  What do you think?
https://github.com/cxd4/AziAudio/commit/bb9d62ff9128746004f8c58166b93ce8d10e81cd

As long as you make sure your xbox_depp.h gets included before Azimer's AudioSpec.h, there's no reason why this shouldn't mangle the spec names for you without a macro function.  (Granted it's less maintainable than your macro function, in case you want to change the prefix at the beginning making each function name unique, but currently this was the only way I figured out to maintain zilmar-spec compatible code in the current codebase while allowing your fork to hopefully prosper.)

So is that sufficient for your needs?  Were there any other issues in other commits here you were thinking about?

weinerschnitzel:
This works. I can handle it from Xbox_depp just fine. Should we move them from the XBox folder to the main source directory? I think some of the source code can get their own folders, too. Maybe for another PR?

cxd4:
Abolish the Xbox folder?

Probably.  There is just that one .cpp and .h file remaining in it, and it does have a more agnostic look to the source code if it doesn't have a unique folder for every portability exception it implements (like just as one example, Xbox).  However I'm not too decided on feeling bothered by the issue into deciding where the files should be moved to, and I don't have a way to test changes (or, without Notepad edit your VS2005 solution to adjust to the moved files/folders), so I feel like leaving that to its own separate PR from this one, if all the other issues for now have been taken care of.

Azimer:
I was hesitant in merging the xbox specific code because it muddled things up so much. This PR has some conflicts I need to resolve so once you feel it's in a good place I will go through it. Thanks for taking a look and trying to make it less xboxy cluttered.

cxd4:
Whoops, conflicts already?

There were no merge conflicts here when I went to bed...somehow one of the other pull requests you just got done merging must have indirectly created a conflict for this one.  Let me try to resolve it...

cxd4:
I'm unsure of the actual cause (not exactly a Git expert), but the conflict seems to be resolved now.

cxd4:
https://github.com/cxd4/AziAudio/commit/601cea5a2af6e6f3f2e40162f04037394ce89a79#commitcomment-11120018

Turned out it was that.  Funny how Git only highlighted the **rest** of the commit from samuelyuan that I merged, when it was really only that one line that was a conflict actually.  Man Git conflicts are confusing.

Azimer:
Going to automerge while I can

PR #54, opened 2015-05-09
Add a function to get mp3data values

samuelyuan:
This should clean up some of the initialization code and make it easier to work with.


cxd4:
You do some good work with this type of thing.  Saying repeatedly `*(s16 *)(base + offset)` throughout the C is a very risky thing (what if (base + offset) isn't 16-bit aligned?) and for central optimizations and avoiding undefined behavior after maintenance is a good idea to have to its own function.

samuelyuan:
That's interesting. I've seen things like that repeated in the code, making me think it was copy and pasted for convenience at the time of development. I just have to make sure that while I'm refactoring that I don't introduce regressions unintentionally

PR #55, opened 2015-05-09
Frequency rounding fix

Frank-74:
Changing leeway from 1000 to 10 fixes the clicks.

Instead of removing the frequency rounding, I've changed the leeway to 10 and added all the sample rate frequencies selectable from my soundcard up to 44100.  Just in case some game uses them.

No real noticeable difference between this and no rounding at all though.


AmbientMalice:
I've been testing this commit.

PURE MAGIC.

No more crackling, especially after loading save states. I assume that was the intent. This pretty much resolves all of Project 64's audio woes. I could be wrong, but Donkey Kong 64 doesn't sound as laggy ingame as it did before.

Frank-74:
@AmbientMalice Would you mind testing this branch I have here.  I've just added an experiment but no more time to test it at the moment.  Attempt on more accurate timing, not sure if its better or worse or neither..

https://github.com/Frank-74/AziAudio/tree/No-Frequency-Rounding

Azimer:
I am not convinced we need to do any rounding at all. If FAT=0 is calculated differently from the playback audio plugin you will receive crackle and pops as the buffer underfills 6 samples or so every second.

AmbientMalice:
@Frank-74 I'll likely be away from my PC for the next 24h. But I'll test that branch when I get a chance.

Frank-74:
> I am not convinced we need to do any rounding at all. If FAT=0 is calculated differently from the playback audio plugin you will receive crackle and pops as the buffer underfills 6 samples or so every second.

I'm closing this PR because I myself don't like it, just tried to satisfy others.

Azimer:
Frank, I want to fix your audio issues. I think there are two parts, one fixing the XA2 implementation and the other reverting the frequency rounding. Then we can move forward from there.

Frank-74:
@Azimer Well my other PR reverts the frequency rounding.  > https://github.com/Azimer/AziAudio/pull/56/files

I have a refinement that gets the Dacrate and Frequency closer but haven't included in that PR.  Much the same as the refinement used in AIDacrateChanged of adding 1 to the Dacrate value.  Adding 1 to the Frequency is more accurate according to the maths with sample rates and video_clock.

PR #57, opened 2015-05-09
Added helper functions to EnvMixer

samuelyuan:
Made a small change in each commit. I added a 16-bit and 32-bit version of two functions only because I wasn't sure if it was necessary or not.


Azimer:
Was this tested?  I am a bit leery on the MixVol function and it would require me to do a bunch of regression testing.

samuelyuan:
Not much really. Are the other functions besides MixVol good to go?

Azimer:
This looks good.

PR #58, opened 2015-05-10
Clean up x86-/zilmar-fixed endianness and memory size guarantees.

cxd4:
While it's not the most likely possibility that we will see a emulator using zilmar's plugin specs which supports big-endian memory (PLUGIN_INFO -> MemorySwaped = FALSE)**, the possibility still exists, and I don't think it's readable HLE code to keep saying ^ 1, ^ 2, ^ 3 all the time when none of those addressing XORs have anything to do with N64 hardware or what is really going on in the RSP, just limitations of most current implementations of the spec's memory layout.

**Not only this but I think MemorySwaped = TRUE shouldn't mean it's swapped on a 32-bit boundary as Mr. zilmar suggests; it should just mean whatever endian is native to your client CPU's machine, whether it be 32-bit little-endian, 64-bit little-endian or even big-endian in which case setting this to TRUE or FALSE should mean no difference.

Anyway the fourth commit I got started on changing non-maintainable types (`short`) to `s16`, because a short is not always a 16-bit type...it is required by the C language to only be anything large enough able to encompass the minimum range of values -32768 <= short <= +32767.  So if we were compiling for a CPU where `short` should be 64 bits then this would be broken.  I can take care of repeated uses of `int` in the source (where 32 bits specifically is required) in a later PR.

### DIFF
--- a/AziAudio/ABI3mp3.cpp
+++ b/AziAudio/ABI3mp3.cpp
@@ -215,7 +215,7 @@ void InnerLoop ();
 //The data will be stored in array v (type s32)
 s16 GetData(int address)
 {
-	return *(s16 *)(mp3data + inPtr + (address ^ 2));
+	return *(s16 *)(mp3data + inPtr + HES(address));
 }
 
 
@@ -515,8 +515,8 @@ void InnerLoop () {
 					for (int i = 7; i >= 0; i--) {
 						v2 += CalcDeWindow(addptr, 0x00, offset, 0x00);
 						v4 += CalcDeWindow(addptr, 0x10, offset, 0x08);
-						v6 += CalcDeWindow(addptr, 0x20, offset, 0x20); 
-						v8 += CalcDeWindow(addptr, 0x30, offset, 0x28); 
+						v6 += CalcDeWindow(addptr, 0x20, offset, 0x20);
+						v8 += CalcDeWindow(addptr, 0x30, offset, 0x28);
 						addptr += 2; offset++;
 					}
 					s32 v0 = v2 + v4;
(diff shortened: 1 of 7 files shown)


Azimer:
Was this tested for regressions?  Between this one and the one @samuelyuan posted I will need to test thoroughly.  There will also be some conflicts as you both are modifying EnvMixer.  So I won't be able to merge right away.  Thanks!

cxd4:
Nope not tested yet, need to flash it to my USB to continue playing with this on my laptop.

And yeah I saw he's modifying EnvMixer as well--my hope is that the functions samuelyuan is contributing are on separate lines from the endian macros I've installed (that way there is no merge conflict), but I did not thoroughly compare every line being changed on his end.  Don't worry; just merge his pull request first and if there is a conflict I will take care of it.

Azimer:
I plan to test this today.  I got really busy this weekend with family stuff.  I will do @samuelyuan 's first then this one.

cxd4:
Audio seems fine here with mk64, gauntlets and mario, couldn't hear any audible regressions.

I'm sure if I had just left it at the first 3 commits and saved the 4th for later this assessment on the safety of the changes would have been able to be much more direct.  That, and the merge conflict was that I replaced `short` with `s16` for 2 sets of 5 lines, which samuel replaced with the function call anyway, which I have just addressed in the last commit I pushed just now.  I guess part of me wanted to risk having to experiment and learn more about merge conflicts on Git; the other part is just impatient and gets excited to accomplish more of related types of changes in the same PR.

Azimer:
Looks good.  Thanks!

cxd4:
Damn, sorry I thought that by selecting the merge-conflict portion of the merge and committing that it would merge it in with the rest of the stuff that wasn't causing the conflict.  I had it backwards--by selectively committing the part that I overwrote to fix the merge conflict, I excluded all the other commits that were contributed and pulled by Azimer since before basing this pull request on.  Sorry about this everyone; if nobody beats me to it by tomorrow I will have time to amend that.

PR #62, opened 2015-05-12
folded almost the entirety of the array multiplies/adds in ABI_Filters

cxd4:
I ran these changes quickly by a full race in _Mario Kart 64_ and _Legend of Zelda:  Ocarina of Time_ and have yet to hear any audible differences, surprisingly.  I do find that surprising, since this is my very first time I have written SIMD using the unusual _mm_madd_epi16 intrinsic:  https://msdn.microsoft.com/en-us/library/yht36sa6%28v=vs.90%29.aspx
(See the fourth commit here for its installation.)

That is where the biggest performance gain comes in.  I'm not sure how much of a bottleneck this portion of RSP audio HLE is, but this function is VERY much faster without repeated scalar multiplies, followed by repeated scalar adds, within a function called 8 times, within a loop repeated at least 16 times.  Makes a big difference.  In fact, I see no remaining ways to increase the performance of this portion of ABI2 emulation that are of relative significance to the rest of the source.

### DIFF
--- a/AziAudio/ABI_Filters.cpp
+++ b/AziAudio/ABI_Filters.cpp
@@ -17,21 +17,12 @@
  */
 static void packed_multiply_accumulate(pi32 acc, pi16 vs, pi16 vt, int offset)
 {
-	i16 source[16], target[8];
 	i32 result;
 	register int i;
 
-	for (i = 0; i < 16; i++)
-		source[i] = vs[i];
-	for (i = 0; i < 8; i++)
-		target[i] = vt[i];
-	swap_elements(&source[0]);
-	swap_elements(&source[8]);
-	swap_elements(&target[0]);
-
 	result = 0;
 	for (i = 0; i < 8; i++)
-		result += (s32)source[i + offset] * (s32)target[i];
+		result += (s32)vs[i + offset] * (s32)vt[i];
 	*(acc) = result;
 	return;
 }
@@ -82,8 +73,10 @@ void FILTER2() {
  */
 	for (i = 0; i < 8; i++)
 		inputs_matrix[15 - (i + 0)] = inp1[i];
+	swap_elements(&inputs_matrix[8]);
 	for (i = 0; i < 8; i++)
 		inputs_matrix[15 - (i + 8)] = inp2[i];
+	swap_elements(&inputs_matrix[0]);
 
 	for (x = 0; x < cnt; x += 0x10) {
 		packed_multiply_accumulate(&out1[0], &inputs_matrix[0], &lutt6[0], 6);
@@ -115,6 +108,8 @@ void FILTER2() {
 
 		for (i = 0; i < 16; i++)
 			inputs_matrix[15 - i] = inp1[i];
+		swap_elements(&inputs_matrix[0]);
+		swap_elements(&inputs_matrix[8]);
 	}
 	//			memcpy (rdram+(t9&0xFFFFFF), dmem+0xFB0, 0x20);
 	memcpy(save, inp2 - 8, 0x10);
--- a/AziAudio/ABI_Filters.cpp
+++ b/AziAudio/ABI_Filters.cpp
@@ -17,10 +17,10 @@
  */
 static void packed_multiply_accumulate(i32 * acc, i16 * vs, i16 * vt)
 {
+	i32 pre_buffer[8];
 	i32 result;
 #ifdef SSE2_SUPPORT
 	__m128i xmm_source, xmm_target;
-	i32 pre_buffer[4];
 
 	xmm_source = _mm_loadu_si128((__m128i *)vs);
 	xmm_target = _mm_loadu_si128((__m128i *)vt);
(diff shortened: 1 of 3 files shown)


cxd4:
Actually in case of non-SSE (e.g. Neon) etc. vector units (or no vector unit at all if your CPU is really old), I just remembered there is a fifth commit I have just now pushed for packing the multiplies and the adds to separate loops so that there is a better chance of auto-vectorization.

Azimer:
I will QA this tomorrow morning I think.  I look forward to seeing more optimization PRs.

cxd4:
Yep there is a  similar optimization to ADPCM, except with nothing to multiply:
```c
	for (u32 x = 0; x < ((k0 & 0xffff) >> 0x4); x++) {
		adpcmtable[0x1 + (x << 3)] = table[0];
		adpcmtable[0x0 + (x << 3)] = table[1];

		adpcmtable[0x3 + (x << 3)] = table[2];
		adpcmtable[0x2 + (x << 3)] = table[3];

		adpcmtable[0x5 + (x << 3)] = table[4];
		adpcmtable[0x4 + (x << 3)] = table[5];

		adpcmtable[0x7 + (x << 3)] = table[6];
		adpcmtable[0x6 + (x << 3)] = table[7];
		table += 8;
	}
```
By rewriting this with swapped elements (i.e. table[6 ^ 1] and table[7 ^ 1], not table[7] and table[6]) we can go straight to an instant 128-bit move.

But for that I plan to wait for this pull request to be processed first, because it defines my `swap_elements` function needed to swap the 8 16-bit halfwords within 4 32-bit words in one instruction.

Azimer:
This is looks good and will be merged.  I will want to discuss the swap_elements at some point to see what you mean exactly by your code comment and how we can improve it in the long term.  
```
/*
 * Unfortunately, as most of the low-level details of RSP hardware came from
 * the ultra-little-endian design from zilmar's RSP interpreter, much of RSP
 * vector simulation in this HLE revolved around backwards element order.
 *
 * A quick fix to this is to have a function to switch the positions of
 * adjacent 16-bit RSP vector elements with each other.
 */
```

cxd4:
In the future, my comment may need to be expanded upon.  I don't like to talk too much, but I don't like to talk too little either (the end result usually being that I still write a lot).

For now I can say this much if it helps you contemplates your questions better.

We all know that the spec most of the time assumes the usual little-endian 32 bits memory layout in RDRAM, DMEM etc.

Well that's not the only thing.  In the RSP interpreter by zilmar, vector registers (local within the RSP plugin themselves) are stored in total little-endian.

What I mean is this.

Literal RSP vector register storage on the native MIPS machine looks like:
```
i16 VR[32][8]; each VR in 0..31 stored as:
{0, 1}, {2, 3}, {4, 5}, {6, 7}, {8, 9}, {A, B}, {C, D}, {E, F}
```
Now since operating directly on each of the 8 16-bit elements on our Intel processors requires byte-swapping, my RSP plugin is forced to store vectors like this, lest we have a huge drop in performance:
```
{1, 0}, {3, 2}, {5, 4}, {7, 6}, {9, 8}, {B, A}, {D, C}, {F, E}
```
The advantage to this method is that I still have the 8 16-bit elements themselves stored in big-endian, like the native hardware does, though the 8-bit byte halves of each element is swapped to little-endian so that we're able to do direct operations on 16-bit inputs.

Now, with the Project64 RSP, we have something even more different:
```
{F, E}, {D, C}, {B, A}, {9, 8}, {7, 6}, {5, 4}, {3, 2}, {1, 0}
```
One big advantage zilmar does gain from this method:  We can operate on 32-, 64-, or even the whole 128-bit segment as an Intel-native integer of that size.  The disadvantage however--the vector unit only ever accesses on intervals of 16 bits; the only time operating on 32- or 64-bit pairs becomes relevant is with the loads and stores under LWC2 or SWC2, which are commonly unaligned anyway which makes that advantage no longer usable.

As a consequence of the method zilmar used, you may see vector element loops like (7 - el) or (15 - byte_element) a lot, when really it could have just been el directly.  I believe this difference in tactical approaches at speed vs. accuracy in our interpreters may amount to part of the reason why you originally had an inverted element order from those in custom HLE arrays you were using, such as what I changed in this commit:
https://github.com/cxd4/AziAudio/commit/c0709c14c53fb52e9525108b05e785362330b659

Here in this commit I changed the left-side of the array assignments to go from 0 to 15 (to match the big-endian order of the source array indexes to the right), instead of 15 to 0 like it started out in your original code.  I believe things such as this may be likely artifacts of the byte-order hacking zilmar did.

Azimer:
I understand this but I wonder if there's a better way since this is just HLE code and the vectors are not persistent across the HLE functions at all.

cxd4:
Fairly sure that there are better ways.  The challenge is looking for it--when I change the byte/halfword order for one single source file accessing it then we have to check all references to the same memory in other source files--offhand I know it's easy to do this with things like the "Find All References" feature in the VS IDE though.

In the meantime I'll see about re-phrasing my comment from back there to yield a less fixed-sounding situation; it makes it sound a little bit like there's a nearly permanent dependency somewhere when there isn't.  However the byte order correction is a relatively insignificant thing to performance, compared to this pull request, as long as it happens externally to the center of work going on deep in the loops.  But in the future yes, I think it would be valuable to re-do some of the endian for some of these arrays between the files which  will eliminate a bit more redundancy.  Meanwhile the `element_swap` function is a faster but still temporary workaround for that to help vectorize the actual operations right away.

PR #64, opened 2015-05-13
Refer to RSP hardware operation `SH` when storing halfwords.

cxd4:
Thanks to samuelyuan we now have already a functional allocation of the repeated behavior of storing the MP3 data as 16-bit segments.  However, the function bears such a striking resemblance to the SH op-code on the RSP, that I felt like re-doing this function to match the syntax of the SH instruction.

SH (store halfword) has the RSP assembly language syntax:
```asm
SH      $rt, +/-offset($base)
```
Naturally since SP DMEM has only 4096 (0xFFF + 1) bytes, whether the offset is positive signed or negative signed isn't exactly significant (could be either an offset of -1 or 0xFFFF, or just 0xFFF for +4095 and still end up meaning the same operation); however that doesn't change the fact that it's still a s16.  I wanted to emulate this in the new function.

As stated in a commit I've tested these changes with Conker's Bad Fur Day...which seems to only ever do aligned SH, not to mention the interesting coincidence that (base + offset) never sets bits outside of ~0x00000FFF which would have caused an access violation for out-of-bounds memory in Azimer's plugin.  So interestingly, the assert's were never needed in the game and I removed them.  Do feel free please to do the usual regression-testing as necessary in case there are other games/cases I don't know about.

### DIFF
--- a/AziAudio/ABI3mp3.cpp
+++ b/AziAudio/ABI3mp3.cpp
@@ -198,11 +198,24 @@ int CalcDeWindow(u32 addptr, int mp3DataIndex, u32 offset, int offsetValue)
 }
 
 // ** Store v[vIndex] -> (mp3DataIndex)**
-void StoreData(s32 vIndex, int mp3DataIndex)
+static void rsp_SH(int rt, s16 offset, u32 base)
 {
-	*(s16 *)(mp3data + mp3DataIndex) = (s16)v[vIndex];
-}
+    u8 * DMEM;
+    u32 address;
 
+    address = base + offset; /* u32 base == GPR[base] */
+#if 1
+    assert((address & ~0xFFFul) == 0);
+#else
+    address = address & 0x00000FFFu;
+#endif
+
+    DMEM = &mp3data[0]; /* mp3data[] has exactly 4096 bytes allocated. */
+#if 1
+    assert((address & 1) == 0);
+#endif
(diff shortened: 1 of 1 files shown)


LegendOfDragoon:
I've never seen a game where SH wasn't aligned. In fact, in PJ64's RSP recompiler, it has a breakpoint to notify that SH is unaligned.

cxd4:
Non-alignment is very common, just not with select operations.

In fact, it doesn't need to be aligned.  If (addr & 3) is 0, 1, or 2, you can still write 16 bits at once even with the DCBA-HGFE memory format the spec does.  So it's not really a question of whether it's aligned, just if it crosses over the 32-bit memory segment barrier OR if addr == 0xFFF, in which case you need to emulate SH as follows.
```c
void SH(int rt, int16_t offset, int base)
{ /* Assume offset == 0x0FFF, and SR[base] is 0. */
    const uint32_t addr = (SR[base] + offset) & 0x00000FFFul;

    DMEM[0xFFF] = (uint8_t)((SR[rt] & 0x0000FF00ul) >> 8);
    DMEM[0x000] = (uint8_t)((SR[rt] & 0x000000FFul) >> 0);
}
```
So aside from that sort of thing, alignment by itself isn't the issue, at least when compared to the high-level SH that Azimer had been doing already in this file.

PR #65, opened 2015-05-13
Fixed typos in comments

dartraiden:


cxd4:
Heh.  70% of these typo fixes are to the engrish zilmar wrote in his audio plugin spec files, which was one of the things I took the liberty of overhauling while re-doing the spec files for cross-platform design.

PR #70, opened 2015-05-14
Simplify ENVMIXER2

samuelyuan:
There's a lot of multiplication, bitshifting, addition, etc. going on in that function. With so many buffers being used, it definitely would help to use a function.

### DIFF
--- a/AziAudio/ABI_Envmixer.cpp
+++ b/AziAudio/ABI_Envmixer.cpp
@@ -108,6 +108,13 @@ void SaveMixer32(int offset, s32 value)
 	*(s32 *)(hleMixerWorkArea + offset) = value;
 }
 
+void buffValueIncr(int x, s16 vec, s16 *buff)
+{
+	int temp = buff[MES(x)] + vec;
+	temp = pack_signed(temp);
+	buff[MES(x)] = temp;
+}
+
 void ENVMIXER() {
 	//static int envmixcnt = 0;
 	u8 flags = (u8)((k0 >> 16) & 0xff);
@@ -366,33 +373,20 @@ void ENVMIXER2() {
 	}
 
 	while (count > 0) {
-		int temp;
 		for (x = 0; x < 0x8; x++) {
 			vec9  = (s16)(((s32)buffs3[MES(x)] * (u32)env[0]) >> 0x10) ^ v2[0];
 			vec10 = (s16)(((s32)buffs3[MES(x)] * (u32)env[2]) >> 0x10) ^ v2[1];
-			temp = bufft6[MES(x)] + vec9;
-			temp = pack_signed(temp);
-			bufft6[MES(x)] = temp;
-			temp = bufft7[MES(x)] + vec10;
-			temp = pack_signed(temp);
-			bufft7[MES(x)] = temp;
+			buffValueIncr(x, vec9, bufft6);
+			buffValueIncr(x, vec10, bufft7);
 			vec9 = (s16)(((s32)vec9  * (u32)env[4]) >> 0x10) ^ v2[2];
 			vec10 = (s16)(((s32)vec10 * (u32)env[4]) >> 0x10) ^ v2[3];
 			if (k0 & 0x10) {
-				temp = buffs0[MES(x)] + vec10;
-				temp = pack_signed(temp);
-				buffs0[MES(x)] = temp;
-				temp = buffs1[MES(x)] + vec9;
-				temp = pack_signed(temp);
-				buffs1[MES(x)] = temp;
+				buffValueIncr(x, vec10, buffs0);
+				buffValueIncr(x, vec9, buffs1);
 			}
 			else {
-				temp = buffs0[MES(x)] + vec9;
-				temp = pack_signed(temp);
-				buffs0[MES(x)] = temp;
-				temp = buffs1[MES(x)] + vec10;
-				temp = pack_signed(temp);
-				buffs1[MES(x)] = temp;
+				buffValueIncr(x, vec9, buffs0);
+				buffValueIncr(x, vec10, buffs1);
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


cxd4:
Looks alright to me, I think it may be more readable, one thing I just noticed with original code ...

Azimer:
Why do we need https://github.com/samuelyuan/AziAudio/commit/7b177eaba0a7e56f8842f3981709f399521afb2b and the temp variables?  Why can't we just go 
```
buff[MES(x)] = pack_signed(buff[MES(x)] + vec);
```
as you suggested.

cxd4:
Yes for that I think just the pack_signed version looks better.  It was probably not easy to see at first due to the existence of a `temp` variable in the first place which gave off the impression it was something more complicated than it really is, hence the instinct to create a new function.  Back before I introduced pack_signed it was just if/else < -32768 / > +32767 which probably did require temp at the time.

In the end, you can never really go wrong with making new functions--in terms of maintainability--but with readability and performance there's always a balance.  Sometimes the lower-level functions or even no functions at all (like a = b + c, not a new function just for adding integers) is more readable and faster.  I'd say with creating new functions, the priority there goes to code that is reused the most number of times throughout the source, but technically even if it's only used 2 times throughout the source, a new function still does improve ease of maintenance somewhat--so it's not really "bad" to always increase the number of functions either.

The more function-based source is written, the more mathematically interesting it is to read and compose, although also the more difficult it can be to navigate what is really being done.

samuelyuan:
Seeing cxd4's comment, I too think that the statement can be compressed. At first, I was trying to make a function out of the block of code, but now I see that it can be improved upon.

Would you suggest I do it in this PR or do it another PR?

cxd4:
If it's convenient enough for you it might be better to append a fourth commit to extend the modification you did in the first commit to compress it down from the function to those statements.  At least that's probably what I would do.

samuelyuan:
Just pushed a new commit. I think it should be all good now.

cxd4:
Any preference to keep/remove the `void buffValueIncr(int x, s16 vec, s16 *buff)` function since it is now a call to a single-line function call or do we prefer to keep it for the sake of having the function

Azimer:
We seem to be sacrificing cpu cycles to make function calls that don't really improve readability or maintainability.  Much of the work will be revamped and rewritten when optimized.  If anything to improve both, the channels should be split up into functions comprised of multiple steps not splitting out the multiple steps into something that feels less readable.  I really think we need to have a good debate on the pros and cons.

cxd4:
When I refer to maintainability, I mean not so much the ability to directly change or read what is happening on the lowest level but to quickly adjust the code as a whole.  Without functions, this requires sort of copy-pasting adjustments throughout, and that is what I mean by maintainability.

Still, in this case, the `buffValueIncr` function ends up just, calling another function (pack_signed).  Best time to make new functions is for volatile code--parts of the HLE that there is reason to think it is likely to need changing, or if it has multiple approaches or algorithms at solving the same problem.  With the pack_signed function taking care of part of the matter, that really just leaves `buff[MES(x)] = pack_signed(buff[MES(x)] + vec);` or in other words just a simple addition on top of the function solving the problem, is on other words no I don't think that's an ideal example for creating a new function.

All of this is ignoring the issue of speed/performance or sacrificing CPU cycles, which is also something that can happen if the compiler isn't aggressive at auto-inlining small functions (most compilers seem to be, but MSVC not so much unless you tell the project settings to inline "wherever suitable").  Like you said, in the long run there will be major changes anyway to revamp and optimize a lot of this, which is why code is better starting off slow and easy to read or update, to end up optimized later on.  (It's unfortunate but the most optimized code isn't always the most beautiful, and I think readability is at least as important for all parties involved.)

Azimer:
It's more than that.  I don't like where some of these changes are going.  For instance, I'd rather have the HleWork area recreated as an 80 byte structure with a Save/Load function than what is presented now.  The single load/save 16/32bit functions do not declutter the code.  MixVol function has misleading parameter names.  

I just think we need to hold back and think through these code changes and really consider what is maintainable and what is just putting things in functions for the sake of doing so.  I will test this merge and commit but I know I am going to have to come back to it later and revise.

On a side note -- I have an idea about the swapping vectors.  Because things end up being 32bit swapped in rdram, 0123 4567 89AB CDEF ends up being 6745 2301 EFCD AB89 requiring the ^1.  The uCode for non-MusyX games use a load/save buffer transfer from rdram.  What if we change it from 32bit swap to 16bit swap under the assumption... everything will be 16bit aligned audio samples.  So we would store it was 2301 6745 AB89 EFCD.  So if we need 32bit access into RDRAM we'll need to do something like *(u16*)(address) | *(u16 *)(address+2) << 16 and 8bit access would be ^1.  I don't recall doing much 32bit or 8bit accesses HLE code.  thoughts?

cxd4:
>For instance, I'd rather have the HleWork area recreated as an 80 byte structure with a Save/Load function than what is presented now. The single load/save 16/32bit functions do not declutter the code. MixVol function has misleading parameter names.

Yeah that's the other issue with creating functions--they should be named based on knowledge of what is intended in the documented HLE solution (along with good function parameter names).  I'm not trying to belittle anything being done here or there, but yeah usually I can see that some created functions have generic names, generic param names, where there was probably an alternative solution away from having to create small/one-step functions to begin with--in the end it turns out easy to avoid functions and use faster solutions.  So yeah I understand what you're saying there.

>Because things end up being 32bit swapped in rdram, 0123 4567 89AB CDEF ends up being 6745 2301 EFCD AB89 requiring the ^1.

At first your example tricked me--I think by "0123" you seem to mean 16 bits:  0x0123.  (I was thinking 0, 1, 2, ... F meant the 16 bytes of a 128-bit vector register; instead yours was 64 bits.)

>What if we change it from 32bit swap to 16bit swap under the assumption... everything will be 16bit aligned audio samples. So we would store it was 2301 6745 AB89 EFCD. So if we need 32bit access into RDRAM we'll need to do something like * (u16*)(address) | *(u16 *)(address+2) << 16 and 8bit access would be ^1. I don't recall doing much 32bit or 8bit accesses HLE code. thoughts?

32-bit access optimization is not so important--16 bits is the native word size for both the SSE2 xmm elements as well as RSP vector element slices.  If you recall little to no 32-bit access, then I say go for it.

That at least is based on my interpretation of your question so far.  (I'm more of a visual learner from examples I think.)  However I have to say that is how I store vectors in my RSP:  Like you said 8-bit access requires ^ 1 to swap both 8-bit halves of the 16-bit little-endian storage, and 32-bit access requires swapping both 16-bit halves to achieve 32-bit Intel endian (like you said with the *(u16 *)address example).

samuelyuan:
I'll admit I do not fully understand a lot of this code in depth. I might have understand it on the surface, but not be able to truly explain it in detail. This might be why some of the function parameters are generic and maybe why some functions don't help as much as they should. Even then, that is no excuse for not attempting to understand the code further. 

With that said, there's definitely a better way to improve upon the code. A function call  for <code>buffValueIncr</code> in this case might not be necessary, although it *might* make refactoring easier. Then again, I think this needs to be put in perspective instead of looked at in isolation.

Azimer:
Thanks for the contribution.  I am going to merge but I may refactor sooner than later to make more sense of the changes.

cxd4:
Just a personal measurement of mine, but the way it looks when it helps code be refactored the most, is when a function centralizes all the possible ways to optimize the plugin for performance.

For example there are at least 3 methods of storing 128 bits:
* XOR the source address by 2 iteratively while moving each halfword.
* Pre-swap the source elements to a temporary vector, then memcpy them over.
* Do an aligned *(__m128 *)d = *(__m128 *)s or something, then swap.
* Refactor functions elsewhere to not adjust the 16-bit endian, then do no swapping at all.
* Consider an SSSE3 solution?
* ...
When there are multiple ways to phrase the same algorithm, then a function helps maintainability.

It's just that for things like
```c
buff[MES(x)] = pack_signed(buff[MES(x)] + vec);
```
then it becomes hard to imagine other possibilities of writing the same algorithm, as it's just a v = func(v + w); or a simple addition, which is hard to optimize/refactor other than the endian issue going on there.  So yeah it depends, but in a fundamental sense there's no harm with making new functions temporarily.

PR #71, opened 2015-05-14
update for ANSI and VS2008 build modes

cxd4:
Partly to fix a bunch of warnings I introduced a while back (saying 0xFFE0 instead of -16 gets implicitly truncated to an s16 by means of truncation), primarily to fix an error from a regression I did in improving the `swap_elements` function.


cxd4:
I did not touch the solution file itself within this pull request.

The vs2008.sln on my disk is stored as Windows CR-LF line breaks; it's stored as Unix LF on your end because you download the raw Git repositories directly off the GitHub website with no conversion of the newlines.  Normally Git processes the line breaks and pulls them to your disk in whatever format is native to your operating system by default (unless you deactivate this feature which is not recommended)--for example all the text files in Azimer's source are stored in LF Unix line endings on the Git machine's side (which is accurate since Git uses a Linux server of some kind I think), but when I execute `git pull` or `git clone` to pull updates from Azimer's online repository to my hard disk, then it is stored to whatever Windows natively uses for line breaks.

So the short answer to your dilemma is that it is a shame that you don't have Git bash/terminal installed. :( It would make your life with GitHub so much easier if you did, than having to download and extract zips all the time.

cxd4:
https://help.github.com/articles/windows-xp-is-not-supported/

That is unfortunate.  I really hope you can download an older version of Git terminal somewhere where it still supported Windows XP.

Frank-74:
Well the Xbx.sln is recognised.  Its in format UNIX not U8-UNIX.

cxd4:
>Well the Xbx.sln is recognised. Its in format UNIX not U8-UNIX.

If, in a hex editor, you delete the 3 byte-order mark bytes at the start of `AziAudio.vs2008.sln`:
`EF BB BF`
Then does that fix it then?  Because I don't mind getting rid of the BOM for UTF or whatever that is.  UTF is dumb Git files should be ASCII. :p

Frank-74:
Nope, that didn't work.  For some reason UNIX format doesn't work for AziAudio either, but it does for the Xbx?  Strange...  I changed to plain DOS format, and had to remove the first blank line before it was recognised.  With U8-DOS the blank line is fine.  It worked before in plain UNIX format.  Grrrr..

Issue #66, opened 2015-05-13
Can `vscale' be shifted by a negative amount?


cxd4:
During one of my last PRs I noticed this line 3 times throughout ADPCM:
https://github.com/cxd4/AziAudio/blob/e189652ff03899f954278a6ab79c9ea598dede5a/AziAudio/ABI_Adpcm.cpp#L89

The comment below it reads:
```c
// so this appears to be a fractional scale based
// on the 12 based inverse of the scale value. note
// that this could be negative, in which case we do
// not use the calculated vscale value... see the
// if(code>12) check below
```
This comment leads me to consider two questions.
* Where is this `if (code > 12)` check?  There doesn't seem to be one throughout the file, which means that if `code` was 12, it doesn't seem prevented that we might be shifting right by a negative.
* Why is it **12**, exactly?  The actual shift amount here reads `((12 - code) - 1)`, which simplifies algebraically to `(11 - code)`; therefore the only ways through which the shift amount would not be negative is `if (code > 11)`, not `if (code > 12)`, is false.

Azimer:
I think that comment came from rcp.  I reversed adpcm myself but reused much of his function.  The code works without issue within the subset of commercial game images so I never paid much attention to it.

cxd4:
Hmm.  How about, keep the first sentence of rcp's comment, discard the note about negative shift amounts at the end, and throw in an `assert(code < 12)`?

Azimer:
The check is in there already but perhaps an assert would be better?

https://github.com/cxd4/AziAudio/blob/e189652ff03899f954278a6ab79c9ea598dede5a/AziAudio/ABI_Adpcm.cpp#L19

cxd4:
That explains it!  So the `vscale` calculation is done only once (per iteration of the `while` loop) for performance, but it is only ever read as a source variable within the `InitInput` functions and that like you said is where the (code < 12) / etc. stuff is managed.

I'm thinking about making a readability change to drop the final parameter to that `InitInput` function you linked, and have `vscale` as a local variable that gets defined within the function itself.  That way the calculation of vscale is in the same block of code as whatever checks (if any) to prevent shifting by a negative number, instead of allowing a shift by a negative number and then undoing it mid-way through a later function call.  Sounds like it could be worth it?

Azimer:
I think a better option is to remove code < srange and the respective parameters and put that assert immediately after code is shifted.  assert(code < 12);  There's no reason to recheck this condition except when code is changed.

[...]

PR #72, opened 2015-05-15
some slight readability changes to ADPCM unit

cxd4:
Further questions I had with the source discussed already at:  https://github.com/Azimer/AziAudio/issues/66#issuecomment-102144485

The last commit is my next question.  Read from un-mapped RDRAM addresses seems to result in reading an unconditional 0, so if we did `addr & 0x7FFFFF` we would break this hardware-accurate behavior just to prevent reading from a 16-megabyte address when only 8 MB RDRAM exists.  So I think it is better to leave the address unmasked hence the last commit, or could this be something to assert?


Azimer:
I didn't test but visually it all looks pretty good!  Hoping I get some time this weekend to put in my contributions.

PR #73, opened 2015-05-16
Poperty Sheets for Visual Studio

project64:
Add property sheets so settings are the same for all, clean up the Vs2008 and Vs2103 solution files


AziAudio/Mupen64plusHLE/types.h
```diff
@@ -0,0 +1,28 @@
...
+#define INT16_MIN        (-32767i16 - 1)
+#define INT16_MAX        32767i16
+
+typedef int32_t 	int_fast32_t;
+
+#else
+#include <stdint.h>
+#endif
+
+#endif
```
cxd4: Shouldn't need this file anymore btw; the compiler dynamics of how/if `<stdint.h>` types exist is all centralized to another header I added earlier to remove the C99 dependency.

Also I'm guessing the primary intention of this PR is the 4th and 5th commits--the 3rd commit you pulled from upstream, the 1st and 2nd commits were changes you did before pulling from upstream which other than changes like this got contradicted by the 3rd commit.

Azimer:
Yeah I am going to have to wade through this one prior to merging.  In addition to 4th and 5th, I believe adding the headers and libs to the 3rdparty folder is intentional.  I am concerned though adding the xaudio2.h file to the 3rd party folder without looking at it.  I believe adding xaudio2.lib from the Windows SDK will cause it to link to XAudio2 2.8 which is incompatible with Windows XP, Vista or 7.

cxd4:
He did add the headers to the 3rd Party folder, but then in his third commit he pulled from upstream where I already added the MinGW-compatible headers to that folder myself, which seems to have constituted a "merge" (in other words he added the headers, but in the end it committed my "changes" to them because I already added the headers myself).

So that leaves the .lib files...which I don't know why are necessary.  Even with MinGW the only unique .lib file I needed to link with it was dsound.lib and only if compiling without XAudio.

project64:
The only reason I added them was to get it to compile, interesting when I tried to compile in VS2008 it did not complain about XAudio2 link but in VS 2013 it did. 

Granted the machine I compiling on does not have any dx or windows sdk installed,  I am installing them to a VMWare machine and just grabbing the necessary files. 

if you have a better more compatible XAudio2.lib file which will work across multiple OS then that is better. 

I also find adding the 3rd party libs means when you build you know it is consistent across different machines, just in case people have different versions of the SDK/DDK's installed.

cxd4:
haven't yet tried pulling from your version, but I edited this in VS 2013 linker Input settings:

`Additional Dependencies:  dsound.lib;kernel32.lib;msvcrt.lib;msvcprt.lib;user32.lib;ole32.lib`
`Ignore All Default Libraries:  Yes (/NODEFAULTLIB)`
and it seems to not complain about missing XAudio2.lib in either Win32/x64 Release builds?

After all I was worried I needed to include an XAudio2.lib for MinGW to get the XAudio2 version to link but if even a non-Microsoft compiler is able to build without the XAudio link-time library then I wonder that Microsoft's official buildchain would require it.  If I really needed it I probably would have included the .lib in the GitHub myself, but I'd really hate to be forced to do that because GitHub should be about text files (better yet text files that are key to the project's sources) rather than binary files.

cxd4:
Just did a clone of project64/AziAudio and indeed the VS2013 solution complains if xaudio2.lib excluded.

cxd4:
@project64 : Found it.
Seems it's because you didn't have an Additional Include Directory specified.

If you specify the additional include directory to your 2013 solution settings to point to your "3rd Party" include folder, you do not need to have XAudio2.lib installed anywhere for the linker.  The absence of this extra include dir seems to cause VS2013 to take precedence with an inconsistent version of the XAudio2 libraries against what Azimer is intending to have the plugin built with.

Frank-74:
Only XAudio 2.8 uses the XAudio2.lib.  Found this link which seems to be applicable to the problem we have supporting old and new OS's.  http://blogs.msdn.com/b/chuckw/archive/2012/04/02/xaudio2-and-windows-8-consumer-preview.aspx

cxd4:
Hopefully that doesn't mean XAudio 2.7 might not work on Windows 8+?
As long as this plugin doesn't actually have XAudio as a DLL dependency I was hoping that'd be recused.

Also, I'm not trying to convey this as though it's the "right" way to feel--it's just how I would feel personally.  But if I were in zilmar's position I would feel like dividing into smaller pull requests based on the VS solutions Azimer already has working.  For one thing I don't know why SSE2 optimizations were turned off this time--because that was intentional or because the new project files in this PR were created from scratch?  I would feel very afraid to risk doing large pull requests with all the possible questions/steps collided into one larger one.

Azimer:
XAudio 2.7 does work on Windows 8+.  It's going forward 2.8+ will not work on anything lower than Windows 8.  It's a shame too because 2.8 removed the dependency on COM and would fix my CoInitialize crash issue.

Azimer:
@project64 : I am having trouble integrating this as a merge even while rebasing.  I am going to have to do this manually.  What are the big things you had to change?  Basically adding the property sheets, dx library, making project file changes, and renaming the 2013 solution?

project64:
well mostly was just trying to get it it compiling, I also used the property sheets that I use in pj64 and cleaned up the settings so they read it from the property sheet instead of from the project.

Is it the best settings maybe not, but it is the base settings I am using for pj64.

cxd4:
There were no issues compiling the plugin with the original 2008 and 2013 solutions, so basically just property sheets then.  Which I have no knowledge about so have no judgment to pass.

You also set the bobby smiles HLE C sources to compile as C++, I think maybe you did this out of habit because before my commits it was required to compile it as C++ to get bool to work in VS2008, otherwise you maybe did it to keep consistent with the rest of it being compiled as C++ which is fine I don't mind that.  But there is no actual _requirement_ to compile it as C++ if that is what you were thinking, or why would it work with Azimer's native VS2008 solution.

Azimer:
I pulled @project64's work and I am piecing things together now.  I am going to push it to a branch and have him look at it before I commit it to the master.  I tried to pick the last two commits but it won't let me because of the change to mytypes.h.  There are some things I do not like about source control.

Azimer:
I am not able to get this to work 100% yet.  I added the property sheets, removed the forced CPP compilation.  I am getting an error in Project64 and 1964 is giving me no audio on either dsound or xaudio2.  I pushed the repository to another branch so we can work out the details.  We also need to iron out all the warnings.  There are way too many warnings.

Azimer:
Okay.  Final status of the night.  I have everything compiling and working it seems.  There are a metric ton of compiler warnings that I need to go through and see if I can reconcile before committing this branch.  I also am not sure if we want SSE2 off by default.  I guess I need to think long term about this and whether I want to check cpuinfo for capability then dynamically allow manually created SSE2 in HLE or not.  What is the target for Project 64?  Are we intended to run at a specific hardware spec?

project64:
For pj64 core it is fine not to have SSE2 .. maybe it is an issue for glide64. RSP will potentially use MMX function for the recompiler, but then it is looking at what the CU will support. 

I only just recently been looking at 64bit which I will compile in VS2013, for the current released one it is in VS2008 aimed at 32bit I guess it supports windows 2000+. But It could possibly go back as far as windows 95/98, really tho I only care about XP+.

Mostly just was trying to get it to compile more then anything. When I took a clean version it did not compile for me. 

Property sheets offer more value for pj64 as there are 18 projects in the solution and I can make sure all have the same base settings. I did aply it to my branch of your audio plugin, just so I know it would have the same settings as pj64. 

With the warnings I have warning level 4 as my default, when VS when creating solutions is warning level 3. While it may not add much value I like the code to be as clean as possible and that helps. Probably my OCD showing though.

cxd4:
I don't mind warning level 4.  I usually set enable all warnings /Wall or /W4 with MSVC for personal projects of mine, just never took it upon myself to make compiler warning levels a priority for others' projects.

Only thing is, any or whatever source of confusion the warning level could create for people makes itself a separate issue from the property sheets along with whatever else was needed to just get it to compile.  If the only thing being changed here was the warning level there would be pretty much no discussion to have before merging it--all different things for contributing, property sheets, warning level, match lack of SSE2 to Project64, compile C as C++--things that could have had their own organized discussion to their own pull request to minimize any room for chaos.

The lack of SSE2 in the project settings could probably be overlooked and merged when we have a 64-bit plugin system, because 64-bit builds unconditionally compile with SSE2 support in them.  So as long as Azimer is able to release 64-bit builds usable in emulators, it could probably be overlooked to keep SSE2 disabled for 32-bit builds.

>Mostly just was trying to get it to compile more then anything. When I took a clean version it did not compile for me.

It must have compiled for you--it's compiled for everyone else here who took a clean version and ran the build straight from the current VS solution.  If it didn't compile for you it's because you didn't pull changes from upstream first and were using the outdated repository as a base for doing the changes (as evidenced by the third commit of this PR where the pull happened midway).

Azimer:
@project64 @cxd4 : I have this PR in https://github.com/Azimer/AziAudio/pull/77 using zil's property sheets.  Please verify it does everything you intend for it.  I used #pragma disable and fixed variables as necessary to remove all the compiler warnings.

Let's transition our discussion into what can we do to make this work for integration in both systems.

Azimer:
Closed in favor of #77.

cxd4:
>With the warnings I have warning level 4 as my default, when VS when creating solutions is warning level 3. While it may not add much value I like the code to be as clean as possible and that helps. Probably my OCD showing though.

@project64 I don't know about Visual Studio with /W4 or /Wall, but all the warnings in AziAudio are going away now compiling with `g++ -Wall -pedantic -ansi`.  You could probably build AziAudio with /Wall set in Visual Studio.  Don't be surprised if you get flooded with warnings coming from Microsoft's own `<windows.h>`, though.

PR #75, opened 2015-05-17
Enable the negative shift amount assertion in ADPCM3.

cxd4:
I felt like showing a commit for this, because the negative shift amount actually happens in DK64.

It was discussed earlier in this issue:  https://github.com/Azimer/AziAudio/issues/66

... that we should be able to change ...
```c
void InitInput(s32 *inp, int index, u8 icode, u8 mask, u8 shifter, u8 code, u8 srange, int vscale)
{
    inp[index] = (s16)((icode & mask) << shifter);
    if (code < srange)
        inp[index] = (inp[index] * vscale) >> 16;
    else
        int catchme = 1;
}
```
... to ...
```c
void InitInput(s32 *inp, int index, u8 icode, u8 mask, u8 shifter, int vscale)
{
    inp[index] = (s16)((icode & mask) << shifter);
    inp[index] = (inp[index] * vscale) >> 16;
}
```
However it turns out that removing the `if (code < srange)` check from the InitInput function broke ADPCM3 which you can hear by booting Donkey Kong.  I don't know what the intention is for cases where the shift amount could be negative, so I'm merely enabling the assert in this commit until @Azimer might have a closer look at that.

### DIFF
--- a/AziAudio/ABI_Adpcm.cpp
+++ b/AziAudio/ABI_Adpcm.cpp
@@ -322,7 +322,7 @@ void ADPCM3() { // Verified to be 100% Accurate...
 		book1 = (s16 *)&adpcmtable[index];
 		book2 = book1 + 8;
 		code >>= 4;									// upper nibble is scale
-#if 0
+#if 1
 		assert((12 - code) - 1 >= 0);
 #endif
 		vscale = 0x8000u >> ((12 - code) - 1);		// very strange. 0x8000 would be .5 in 16:16 format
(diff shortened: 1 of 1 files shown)


Azimer:
It basically turns it into a noop...  I will review.  We might be able to do a check and instead of asserting set vscale to 0x10000.  In the loop it will do 
```
 		inp[index] = (inp[index] * 0x10000) >> 16;
```
Which is essentially inp = (inp << 16) >> 16;  Doing it this way would remove a jump from the inner loop and treat it always as if it has a valid value.  Thought?

cxd4:
Yeah actually the idea you just said was exactly what I just tried before doing this pull request, except that I did it wrong (I lazily copy-pasta'd the assert condition as the condition of when to set vscale to 0x10000) so we were thinking the same thing there.

However when I do this (see commit) it works:  https://github.com/cxd4/AziAudio/commit/2e0bd60b9bc234f85f4f00f3668aee86f87b61c1
Something like that?

Azimer:
Yep.  That is exactly what I was thinking.

PR #77, opened 2015-05-17
Add Project64 project sheets to ease integration

Azimer:
Adds @Project64's project sheets.


cxd4:
As far as I can tell, it looks good.

Azimer:
The inferred regression in the last commit was actually a regression along the lines of DK64 where negative vscale values are indeed okay.

project64:
when I compile AziAudio.sln:
1>     Creating library D:\My Programs\AziAudio\Debug\AziAudio.lib and object D:\My Programs\AziAudio\Debug\AziAudio.exp
1>XAudio2SoundDriver.obj : error LNK2019: unresolved external symbol __imp__XAudio2Create@12 referenced in function "public: int __thiscall XAudio2SoundDriver::Setup(void)" (?Setup@XAudio2SoundDriver@@QAEHXZ)
1>D:\My Programs\AziAudio\Debug\AziAudio.dll : fatal error LNK1120: 1 unresolved externals

in VS2008 project I get:
1>d:\my programs\aziaudio.5\aziaudio\xaudio2sounddriver.h(16) : fatal error C1083: Cannot open include file: 'xaudio2.h': No such file or directory

Frank-74:
@project64 Have you added the DirectX SDK include paths to Visual Studio?  I had to add them manually.
vs2008include

Azimer:
Hi @project64.  I think you are accidentally pulling master rather than the branch project64-add-project-sheets .  I use Windows 8 and VS2013 and I have no issues.

project64:
@Frank-74 I do not have Direct X sdk installed on this machine, so I do not have it in the VS project settings

cxd4:
I said this already.  You're supposed to use the headers under the 3rd Party folder.

If it's the xaudio2.h and comdecl.h that I included with the repo then it compiles against the XAudio2.7 API which means _it's impossible for you to be getting unresolved XAudio linker references_.

Unless you can confirm that the branch you pulled from Azimer does actually not have the additional include directory specified for the mode being compiled, which it shouldn't and if it doesn't I am positive that's why you get the link failure.

cxd4:
Due to my lack of Git command knowledge I just did:
```
git clone https://github.com/Azimer/AziAudio.git
cd AziAudio
git checkout project64-add-property-sheets
```
then opened the AziAudio.vs2013.sln

then set /NODEFAULTLIB in Linker input settings

link only against, nothing else:
`kernel32.lib;msvcrt.lib;msvcprt.lib;user32.lib;ole32.lib`

I have no problems.
If compiling without XAudio then you also need dsound.lib, but *nobody* should be getting errors pertaining to a missing xaudio.lib, because um, that's not needed.

project64:
@Azimer  you were right, I thought I had switched branches but I had not. 

it seems to compile fine for me with that branch

Azimer:
Very well.  My only concern now is performance.  The release build is 2x slower.  Some of it's SSE2 and some of it's other optimization settings.  I am going to compare the two projects and we can discuss what we're going to do between the two.  These settings might even be considered better new settings for the entire property sheet; if not perhaps just for the audio plugin.  I do believe the default of SSE2 off might be okay.  The way I see it though, we will probably want an SSE2 enabled, disabled, dsound and xaudio2 combinations.  How do we want to present that??  or is it better just to have it dynamic with a larger binary?  Thoughts?

project64:
It will be interesting to see what you change to get the performance to match the previous build, maybe I should use those settings in pj64. 

the question is how many users would not have SSE2 on there machine, most likely either should be on by default, or look at how to make it a dynamic option.

cxd4:
>I do believe the default of SSE2 off might be okay. The way I see it though, we will probably want an SSE2 enabled, disabled, dsound and xaudio2 combinations. How do we want to present that?? or is it better just to have it dynamic with a larger binary?

We need a 64-bit plugin system Azimer.  Once we have that, the assumption of common existence of SSE2 hardware support can be moved to the assumption that most people today have 64-bit CPUs, which guarantees SSE2 availability.  This way we can safely keep SSE2=disabled for the people stubborn enough to still have a 32-bit-only CPU while the 64-bit builds unconditionally compile with 64-bit support in them.

As for DirectSound vs. XAudio, I typically just compile your plugin with XAUDIO_LIBRARIES_UNAVAILABLE defined for a DirectSound build because I think? that has less crackling on my CPU currently, but as the intention is to move to XAudio2 I don't mind if you just do XAudio-only releases to the public.  So I think it's safe to impose XAudio on everyone who's too lazy to build it themselves otherwise without the need to worry about providing a run-time toggleable option for switching which API.

cxd4:
>My only concern now is performance. The release build is 2x slower. Some of it's SSE2 and some of it's other optimization settings.

Yeah it's set for smallest optimizable build size, also functions are no longer getting in-lined like they were before (MSVC never inlines functions unless you tell it to explicitly):
http://ft.trillian.im/785abe041074fd64135ab3318aff1ec8767ab03d/6yBGHFotjgRpj7cLDeFHia2ynhNEy.jpg

Since functions are "called" on a regular basis in the main ABI HLE loops in some of the sources, calling these procedures that many loop iterations begins to take it's toll as well.

cxd4:
@project64 how much damage would it do to just have SSE2 enabled for 32-bit Release builds?

This isn't just a problem for Azimer; it affects you as well.  Currently you have an all-in-one emulators + emulator plugins + optional components sort of feature of releasing.  For example the Project64.exe needs basically nothing from SSE2, because the N64 VR4300 CPU has no SIMD.  However SSE2 is a requirement for accurately using x86 to emulate the RCP, with or without the physical existence of re-compilers.  So for some plugins, you desperately need SSE2 to keep the speed and accuracy; for Project64, you need no SSE at all except maybe for fast emulation of COP1.

The other problem with the property sheets is that they're relevant less to the best settings for the audio plugin, more to the best settings for the CPU core emulator Project64 which has a totally different need of optimization methods than the RCP has, which you bundle under the same property sheets just so Project64 is able to present itself as if there were no plugin holes missing (pre-include personally preferred plugins which require different optimization settings for best results).

The other problem with the property sheets is that there's nothing fundamental about them.  Git repositories can flourish as complex as forks of forks of forks...it can't be freely done if some other emulator has their own PropertySheets and preprocessor settings besides the ones that Project64 uses.  This causes an unbalance because no other project can use property sheets without a) replacing theirs with yours or b) doing a fork of Azimer's repository deleting out the property sheets you have and manually committing theirs in for a repo conflict, when really the only topical part of this repository should be the source to Azimer's AI emulation + HLE (that is why the repo exists).  There may be plenty of people in your same exact position zilmar--how is it fair to all of them to force only them to have to replace their PropertySheets/remove them to get them to compile but not you if everyone is the same?

One solution could be to keep AziAudio.sln; use zilmar's AziAudio.vs2013.sln as the version that uses Project64-configured property sheets.  Also simply removing the property sheets from upstream shouldn't break anything with pulls/merges between the two AziAudio repo's as long as only one person is using/changing the property sheets at all for anything.

If your genuine intention was just to get the plugin to compile successfully then you should have just left your initial pull request at that.

Azimer:
I am going to keep looking at this.  It seems even with propertysheets a project can override those settings by setting its own values.  I need review those settings yet.  I had a busy work day yesterday and an evening with the wife.

cxd4:
Good news, that.

Azimer:
Here are the differences I've uncovered which seem potentially to affect optimization or are otherwise inconvenient.  I am going to give this some time to provoke discussion then tomorrow I am going to set these in the x86 and x64 Release projects and go except for SSE2 (which we can keep discussing)).

| Description | Property Sheets | AziAudio project |
|----|----|----|
| General - Use Link Time Code Generation | Not Set | Set |
| C/C++ Warnings | Level 4 | Level 3 |
| Enable C++ Exceptions | Yes with SEH Exceptions (/EHa) | Yes /EHsc |
| Code Generation | /arch:IA32 | /arch:SSE2 |
|  Instrinsic Functions | No | Yes /Oi |
|  Favor Size Or Speed | Favor small code | Neither |
|  Inline Function Expansion | Default | Anything suitable /Ob2 |
|  Optimization | Minimize Size /O1 | Full Optimization /Ox |
|  Runtime Library | Multi-threaded /MT | Multi-thraded DLL /MTd |
|  Whole Program Optimization | No | Yes /GL |
| Linker - Use link time code generation | Not Set | Link Time Code Generation /LTCG |

Personally, all except for C/C++ Warnings should be changed from my perspective.  I took the time to fix or bypass the warnings within the source code.  /MTd may be something set special within each project.

LegendOfDragoon:
I personally prefer favoring speed over size. Although ideally, it would be nice to be able to select different optimization levels for specific functions. I also think having a higher warning level is sometimes nice. It can aware you of some mistakes.

Azimer:
I don't think it would make much difference to select different optimization levels for specific functions.  What I am foreseeing though is dynamically swapping out HLE functions in the jump table when SSE2 is enabled.  Thinking of which.  I believe we should at least be able to compile as SSE by default.  I don't foresee why a P3-450Mhz could not be a minimum requirement.  At least with SSE, we'd have enough optimization to potentially make the first gen P3 run at full speed in theory.

LegendOfDragoon:
The reason to do it would be to reduce size of the binary output for code that isn't run very often. Not all that important, unless you care about the binary size. I generally just select favor speed. I don't bother with pragma's or anything like that.

@Azimer I think dynamically swapping out HLE functions is a great idea.

cxd4:
_Use Link Time Code Generation_, _Whole Program Optimization_, Use link time code generation_
I could go either way with these on or off...but I'm not sure that LTO really makes that much of a performance gain if the cpp modules are all organized to completely isolated HLE functions (that don't interact much with functions in other sources) so wouldn't really mind keeping LTO off the way the PropertySheets have it.  However this needs testing and not simply my skepticism.

_C/C++ Warnings_
If I could I like "Enable All Warnings", but not if it means using macros/pragma to disable specific warnings or achieving the solution of warnings in a bypass/ignore style of way.  In that sense I can see some pragmas have been used already to solve Level 4 warnings and am not even sure about using that over level 3; however I would love to use level 4 or enable all warnings if it means having an internal fix for every warning and can look into that later.

_Code Generation_
Just be sure to set this to SSE2 only for 32-bit release builds Azimer.  If you leave the x64 build set to SSE2 then MSVC will complain about an unrecognized compiler switch, because SSE2 is guaranteed as a built-in part of x64 64-bit compiling modes.  So we just leave x64 to "Not Set" instead of SSE2 and it will generate the SSE2 anyway with no complaint.

_Runtime Library_
I hate not only to bump up the DLL size but also statically link the extra crap sections generated from having the C/pp runtime stuff statically linked into the frame sections of the image (never minding the increased build size it just adds some ugly bytes imo).  However this may be a necessary evil for compatible releases which work for everybody especially when releasing under Project64 as one of Project64's first-party plugins so in that sense I'm pretty much okay with it.

That's all I can think to comment at this time.

project64:
> Code Generation
the property sheets should have SSE disabled by default, now your project overriding this and having it enabled for your project I am not sure which way to jump there, that needs more discussion, but that can be turned on by default in the project to override the property sheets and we can have a different discussion about that one. 

> Runtime Library
the advantage of Multi-threaded /MT over Multi-thraded DLL /MD is that I do not need to include the VS run time library. It does make the binary bigger, but when you look at the installer size it is actually smaller with the bigger DLL's but no run time.

And I like trying to reduce dependencies

> Favor Size Or Speed
Strange that I actually found the favor size to be faster, granted it probably was not a great sample size and test base to be exact but figured it did not really make much difference which one. I think the reason the smaller produced better speed meant less flushing of caches and more staying in memory

> Enable C++ Exceptions
the "Yes with SEH Exceptions" is important for pj64 since the way I do exception handling with RDRAM. It most likely irrelevant which is the option you end up using for the plugin. But then again I hate using c++ catch/throw mechanism. So that does not matter so much for me.

> C/C++ Warnings
I like level 4 just to have the code as clean as possible, similar to code generation maybe have the project override it back to level 3 and look at future pull request that fixes the warning and changes it to level 4. 

> Inline Function Expansion
Maybe I should have Anything suitable /Ob2 as the default, not sure.

Azimer:
After putting some thought into this (and the fact I need to move forward with this or dump it), I am going to do the following.  I am going to temporary override the properysheets on every setting except warnings, exception, code generation, runtime library.  I think this will allow it to work in PJ64 without any real drawbacks.  I will also consider adding another configuration specifically intended for our testing that keeps our old settings.  This shouldn't interfere with Project 64 since it doesn't use anything except for Debug/Release Win32/x64.

If there are any more concerns speak up before I make the changes, push the commit and merge the PR in about 12-15 hours.

project64:
@Azimer  sounds good, the property sheets are really just meant to be the defaults.

cxd4:
No more concerns from me about the suggestions Azimer sounds good.

I don't even mind much whether the run-time is statically or dynamically linked within AziAudio.dll since both answers have their obvious advantages, but that doesn't mean I'm going to ignore what I perceive to be some technical flaws in part of project64's reasons for it.

>the advantage of Multi-threaded /MT over Multi-thraded DLL /MD is that I do not need to include the VS run time library. It does make the binary bigger, but when you look at the installer size it is actually smaller with the bigger DLL's but no run time.

How can we be sure @project64 ?  You could end up with a complete set of [even more?] plugins and the core, all requiring MSVCR120.  How can you be sure that 8 modules statically linked with the CRT is smaller installer size than 8 executable modules all linked dynamically to the same central DLL?

Even if you're right, couldn't you have just coded the Project64 installer to detect if msvcr??? already exists on the user's system and only **then** install the dll if it doesn't?  Think about this more carefully because most users have already been forced to run updates/get msvcrxxx on their system already--by the time they're through running your installer they have no more need for the installer EXE because Project64 is already installed.  So the installer extracts msvcr120.dll or whatever *only if it can't be found on the user's system already* which is quite unlikely, therefore saving even more bytes from everyone's installation of Project64 if you had just done that.

>And I like trying to reduce dependencies

Dependencies increase build size, not decrease it.
Statically linking the dependencies doesn't eliminate them, just moves where they are stored.

The only way to reduce _dependencies_ is to change the code to not _depend_ on them--e.g. no stdlib.

Azimer:
To get the plugin back to the size/speed we had before, we will need to set SSE2 and change the runtime library from /MT to /MD.  For releases separate of PJ64, I plan to set SSE2 and /MD to for speed and code size but what I've compromised on is sufficient.

Time to merge this and we can deal with any changes afterward.

AziAudio/Mupen64plusHLE/musyx.c
```diff
@@ -845,7 +845,7 @@ static void sfx_stage(struct hle_t* hle, mix_sfx_with_main_subframes_t mix_sfx_w
     mix_fir4(musyx->e50, buffer + 1, fir4_hgain, fir4_hcoeffs);
     dram_store_u16(hle, (uint16_t *)musyx->e50, cbuffer_ptr + pos * 2, SUBFRAME_SIZE);
 }
-
+#pragma warning(disable : 4100)
```
cxd4: This is the only #pragma warning disable & default left in the source I think.

The others were fixed natively.

Just to not interfere more than what would be useful to interfere...I left this in for GCC (which doesn't support #pragma warning disable) but #ifdef'd it under _MSC_VER to fix the unknown #pragma GCC warnings.

PR #78, opened 2015-05-17
Compressed code in EnvMixer

samuelyuan:
I noticed that in the previous pull request #70, we made use of the pack_signed function to reduce a lot of clutter while still maintaining readability (i.e. we turned 3 lines of code into a single line by removing a <code>temp</code> variable)

In the main loop of EnvMixer3(), it is incrementing <code>out[MES(y)], aux1[MES(y)]</code>, etc. I was thinking that instead of having all those temporary variables like <code>o1</code> and <code>a1</code>, why not replace it with one line of addition using the pack_signed function?

### DIFF
--- a/AziAudio/ABI_Envmixer.cpp
+++ b/AziAudio/ABI_Envmixer.cpp
@@ -519,35 +519,18 @@ void ENVMIXER3() {
 		// ****************************************************************
 		MainL = MixVol(Dry, LVol);
 		MainR = MixVol(Dry, RVol);
-
 		i1 = inp[MES(y)];
 
-		o1 = out[MES(y)] + MixVol(i1, MainL);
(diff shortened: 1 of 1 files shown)


cxd4:
Seems good to me otherwise.

In fact from what I've seen compilers seem to pick up on corner-case advantages better when you do this:
```c
x = n[0] + n[1] + n[2] + n[3] + n[4];
```
... rather than this ...
```c
x = 0;
for (i = 0; i < 5; i++)
    x += n[i];
```
So it can help some things to be more optimized.

Azimer:
This is good work.  I will look it over.  Thanks!

Azimer:
You can keep adding to this PR if you want but I am not going to merge it until I finish the major revision sitting in the project64 branch.  Once this is done I will manually merge and fix any conflicts.

PR #80, opened 2015-05-21
Change RomOpened to RomOpen

LegendOfDragoon:
Due to the fact that Project64 2.x uses "RomOpen", the function
"RomOpened" is never used. Now RomOpen works. Hopefully this may help
with tracking down bugs :smile: .

### DIFF
--- a/AziAudio/AudioSpec.h
+++ b/AziAudio/AudioSpec.h
@@ -197,7 +197,7 @@ EXPORT void CALL ProcessAList(void);
   output:   none
 *******************************************************************/ 
 EXPORT void CALL RomClosed(void);
-EXPORT void CALL RomOpened(void);
+EXPORT void CALL RomOpen(void);
 EXPORT void CALL PluginLoaded(void);
 
 EXPORT void CALL AiCallBack(void);
--- a/AziAudio/main.cpp
+++ b/AziAudio/main.cpp
@@ -287,7 +287,7 @@ EXPORT void CALL ProcessAList(void) {
 	}
 }
 
-EXPORT void CALL RomOpened(void) {
+EXPORT void CALL RomOpen(void) {
 	ChangeABI(0);
 	snd.DeInitialize();
 	Dacrate = 0;


Frank-74:
Nice :)

This has made things more consistent here.  Now switching from any plugin to XA2 while a game is running, results in no sound.  End Emulation with XA2 and restart game and sound is back.  Hopefully this is reproducible your end too.  XA2 still causes crashes after a while, but DS8 no longer crashes my PC when closing down PJ64 with a game running.

LegendOfDragoon:
@Frank-74 how about removing snd.Initialize() in RomClosed? I put audioIsInitialized = FALSE; instead :smile: . This doesn't fix the save state issue where there's no sound, but it rarely crashes. Although this makes the plugin not work properly on other emulators.

I personally would avoid using RomOpen, just for the sake of compatibility with other emulators. I just like to experiment with it sometimes.

Frank-74:
Confused about this RomOpen/RomOpened with PJ64.  To me, RomOpened is after initialize, RomOpen is before.  Why was RomOpened removed from PJ64?

Also, this doesn't look right to me.
```cpp
BOOL XAudio2SoundDriver::Setup()
{
	if (dllInitialized == true) return true;
	dllInitialized = true;
```

cxd4:
I don't know if zilmar intended a difference between the two.  But Gfx #1.3, Controller #1.1, Rsp #1.2 ... they all have a history of introducing a new function named `RomOpen` instead of having to use the Initiate* functions (InitiateGFX, InitiateAudio, InitiateControllers ...), so it's inconsistent to suddenly name the function RomOpen**ed** in just one of the specs.

Azimer:
The spec leaves a lot open to interpretation.  I did a lot of things like that to make sure the plugin won't crash.  For example, opening the configuration dialog prior to initializing directsound and enumerating hardware would end in a crash.  But you can't always guarantee Initialize is called prior.

I plan to revamp it all and focus only on PJ64 since it's really the only active emulator using the spec.

theboy181:
Will your focus kill compatibility on other emulators? I know there are a bunch of us who use different emulators for features that are not supported in PJ64.  ie. OC function in 1964..  i would hate to not be able to use the best public sound plugin.   :(

cxd4:
I think he's just talking about subtle issues like threading/timing of initialization/closing of the API.

I know I had a hell of a lot of investigating to do to please every emulator's implementation of the spec with the way that OpenGL context sharing and display lists threading worked...could only be called within certain functions and had to create semaphore variables to hack around emulators that didn't do the spec the same way zilmar did (while complying to its open-ended rules nonetheless).  But none of that kind of thing makes the plugin usable only on one emulator, no, just prone to some effects.

Azimer:
@theboy181 : Not intentionally but I am not going to test any changes made on any other emulator.

Frank-74:
Hmmm, Plugin Class.cpp uses RomOpened.
```cpp
void CPlugins::RomOpened ( void ) 
{
	m_Gfx->RomOpened();
	m_RSP->RomOpened();
	m_Audio->RomOpened();
	m_Control->RomOpened();
}

void CPlugins::RomClosed ( void ) 
{
	m_Gfx->RomClose();
	m_RSP->RomClose();
	m_Audio->RomClose();
	m_Control->RomClose();
}
```

Frank-74:
Removing snd.DeInitialize(); from RomOpen and RomClosed, as well as audioIsInitialized = FALSE; in RomClosed only.  And no more crashing here switching between DS8 and XA2.

LegendOfDragoon:
@Frank-74 , I just tested 1.7 and 2.1. Neither of them use "RomOpened" when booting a game or resetting.

@theboy181 , You'll have to use a different audio plugin, if you use 1964.

PR #81, opened 2015-05-22
Condensed loops in ADPCM

samuelyuan:
Rewriting some of the while loops with for loops makes it clearer that the loop is initializing pairs of inputs, meaning for each even <code>i</code> from 0 to 8, initialize <code>inp[i]</code> and <code>inp[i + 1]</code>.

Also made a minor change where
<code>
a[MES(j)] >>= 11; a[MES(j)] = pack_signed(a[MES(j)]);
</code>
becomes
<code>
a[MES(j)] = pack_signed(a[MES(j) >> 11]);
</code>

### DIFF
--- a/AziAudio/ABI_Adpcm.cpp
+++ b/AziAudio/ABI_Adpcm.cpp
@@ -44,7 +44,6 @@ void ADPCM() { // Work in progress! :)
 	s16 count = (s16)AudioCount;
 	int vscale;
 	WORD index;
-	WORD j;
 	s32 a[8];
 	s16* book1;
 	s16* book2;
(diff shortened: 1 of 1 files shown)


cxd4:
As for using `for` instead of `while`, I'd say mathematically it's more readable in terms of sequences.

This would be the other reasonable way of writing it I think.
Instead of
```c
for (int i = 0; i < 8; i += 2)
{
    InitInput(inp2, i, icode, 0xf0, 8, vscale); // this will in effect be signed
    InitInput(inp2, i + 1, icode, 0xf, 12, vscale);
}
```
... I believe we could do ...
```c
for (i = 0; i < 8; i++)
    InitInput(inp2, i, icode, 0xF0 >> 4*(i % 2), 8 + 4*(i % 2), vscale);
```
However in spite of the slight increase in maintenance this is slightly less likely for the compiler to unroll the loop into the 8 iterations, so it has its advantages, disadvantages.  May have to check the asm output from the compiler to see if it will still expand the loop.  Either way nice job.

PR #83, opened 2015-05-23
vectorized almost the entirety of what's left of ABI_Adpcm.cpp

cxd4:
Okay this one was very challenging and time-consuming for me.  It wasn't the hardest SSE work I've ever done in any plugin, but so far in this plugin it was the most tricky for me to implement.

I say "almost the entirety".  A few things left are like, memcpy...things totally insignificant to gains in performance compared to this function I have touched.

Got to say I was under the pressure of time trying to finish up--do let me know if there is anything to nitpick.  Needless to say it is more complex than for me to just be happy naming a couple games I can regression test, so it can need more testing.

### DIFF
--- a/AziAudio/ABI_Adpcm.cpp
+++ b/AziAudio/ABI_Adpcm.cpp
@@ -22,7 +22,7 @@ void InitInput(s32 *inp, int index, u8 icode, u8 mask, u8 shifter, int vscale)
 void ADPCMFillArray(s32 *a, s16* book1, s16* book2, s16 l1, s16 l2, s32 *inp)
 {
 	s32 b[8];
-	register int i, j;
+	register int i;
 
 	for (i = 0; i < 8; i++)
 		a[i]  = (s32)l1;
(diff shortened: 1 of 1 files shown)


AziAudio/ABI_Adpcm.cpp
```diff
@@ -13,24 +13,83 @@
...
+	xmm_target = _mm_unpacklo_epi16(prod_n, prod_m);
+	xmm_source = _mm_unpackhi_epi16(prod_n, prod_m);
+
+/*
+ *  for (i = 0; i < 8; i++)
+ *      products[i]  = (l1[i] * book1[i]) + (l2[i] * book2[i]);
+ */
+	prod_hi = _mm_add_epi32(prod_hi, xmm_target);
+	prod_lo = _mm_add_epi32(prod_lo, xmm_source);
+
```
cxd4: In between this and the two storeu lines after this, I was going to add this:
```c
/*
 *  for (i = 0; i < 8; i++)
 *      products[i] += 2048 * (s32)((s16)inp[i]);
 */
	xmm_source = _mm_setzero_si128();
	xmm_target = _mm_loadu_si128((__m128i *)&inp[0]);
	prod_m = _mm_unpackhi_epi16(xmm_source, xmm_target);
	prod_m = _mm_slli_epi32(prod_m, 11);
	prod_n = _mm_unpacklo_epi16(xmm_source, xmm_target);
	prod_n = _mm_slli_epi32(prod_n, 11);

	prod_hi = _mm_add_epi32(prod_hi, prod_m);
	prod_lo = _mm_add_epi32(prod_lo, prod_n);
```
However I've tried and tried but some sort of bug persists in that code.  I was unable to isolate the bug in time before my time was up, so I just left it to the ANSI C looped version:
```c
for (i = 0; i < 8; i++)
    a[i] += 2048 * inp[i];
```
Which is 3 SSE instructions more in the compiler output, which isn't too bad a deal.  It's still automatically generating the SSE well enough that I am not worried about it so badly.

PR #86, opened 2015-05-24
completed SSE2 optimizations on new bsmiles code in POLEF()

cxd4:
After I saw the new filter implemented just recently to support some sound effects going on in _PilotWings_ or whatever I immediately saw a couple things to vectorize about it.  However I have done no regression-testing whatsoever, because I don't know of any games that use PilotWings ucode (heard something about Body Harvest, StarFox maybe?  bunch of games that weirded me out never really had any motive to get familiar with them).

In case there is a regression, this is why I do multiple commits.  It's also nice to show my work step-by-step; that was one of my favorite things to do in math classes. :)

### DIFF
--- a/AziAudio/ABI_Filters.cpp
+++ b/AziAudio/ABI_Filters.cpp
@@ -43,14 +43,14 @@ static void packed_multiply_accumulate(i32 * acc, i16 * vs, i16 * vt)
 // rdot is borrowed from Mupen64Plus audio.c file.
 s32 rdot(size_t n, const s16 *x, const s16 *y)
 {
-	s32 accu = 0;
+	register size_t i;
+	s32 accu;
 
 	y += n;
 
-	while (n != 0) {
+	accu = 0;
+	for (i = 0; i < n; i++)
 		accu += *(x++) * *(--y);
-		--n;
-	}
 
 	return accu;
 }
--- a/AziAudio/ABI_Filters.cpp
+++ b/AziAudio/ABI_Filters.cpp
@@ -228,14 +228,44 @@ void POLEF()
 		s32 accumulators[8];
 		s16 frame[8];
 
-		copy_vector(&frame[0], &inp[0]);
+#if defined(SSE2_SUPPORT)
+		xmm_target = _mm_loadu_si128((__m128i *)inp);
+		_mm_storeu_si128((__m128i *)&frame[0], xmm_target);
 
-		for (i = 0; i < 8; ++i)
+		prod_m = _mm_mulhi_epi16(xmm_target, xmm_source);
+		prod_n = _mm_mullo_epi16(xmm_target, xmm_source);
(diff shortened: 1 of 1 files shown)


LegendOfDragoon:
Turok also uses POLEF iirc. PilotWings uses the most common ucode, but it's just that not all games use the POLEF command, or at least is rarely used in some games.

cxd4:
Not familiar with 1/2/3 Turok or whichever I have but don't know how to play; I guess I'll see about putting a MessageBox or an assert() somewhere in POLEF to show me when some games I'm familiar with how to play use it.

Azimer:
I won't be able to commit this until I figure out the issue.  So this may hang out there for a little bit then I will merge and apply the fix.

Azimer:
Should include the games that are having regression / difficulty are Pilotwings (which is fixed) AeroFighters Assault, and GoldenEye.

cxd4:
I can try to look for it, too.  GoldenEye I most certainly know how to test.

Is it just bsmiles `alist_polef()` backport that has the issue?  Or some other functions we're looking at?
https://github.com/mupen64plus/mupen64plus-rsp-hle/blob/master/src/alist.c#L909

Azimer:
The trouble seems only with the backport.  The static is not there when POLEF is disabled even though the audio is wrong.  I know one issue already is the input buffer needs to care about endianess and it's not.  But doing inp[i^1] does not fix anything so I didn't bother to commit a fix yet.  There is a problem elsewhere.  The table I call adpcmtable is always swapped so that should be fine.  It has to be somewhere though...

cxd4:
I can't seem to boot GoldenEye on pj64 with this plugin, keeps complaining a breakpoint in RunRSP.

So I think I'll build a HLE audio version of my RSP so I can test it pj64 1.6.

cxd4:
Doing inp[i ^ 1] actually does seem to fix something...but it's a quieter issue hidden underneath something else just like you said.  I believe I fixed that here in this separate pull request I just did:  https://github.com/cxd4/AziAudio/commit/baee243ba190ce76f5b4b9bf037079b6f3cd08f0

After doing that commit, GE has some quieter fuzzing in the background noise, which I took what you said about inp[i ^ 1] and now that is fixed as well.  I fixed that issue within this pull request (instead of the other one I just linked to above) to avoid having to resolve merge conflicts.  The last commit I just pushed to this PR is the one I mean; it does inp at i^1 in the form of calling swap_elements instead of copy_vector.

So merging these two PRs seems to have flawless audio in GoldenEye now with POLEF in-tact.

However the SSE2 codepath seems to have some loud regressions, so this PR might not be ready for merging yet.  (Then again SSE2 is disabled in the project settings, so if this PR had been merged right now it wouldn't be noticeable yet.)

Azimer:
I am going to merge this and the other one.  We can fix the SSE2 later.

PR #87, opened 2015-05-24
Compress resample

samuelyuan:
Resample hasn't been touched much, so now is a good time to make changes. I applied the pack_signed function as well as a few other small changes. 

Note that most of the calculations take place in <code>MultAddLUT</code>, formerly called IncrAccum. I had to change the function name for IncrAccum because it was potentially misleading, given that there were two variables <code>accum</code> and <code>Accum</code> (just a small difference involving lowercase and uppercase). <code>accum</code> is more of a temp variable, but it has nothing to do with <code>Accum</code>. I renamed it to <code>MultAddLUT</code> since the function was doing multiplication and addition, but feel free to change it to a more appropriate name if needed.

### DIFF
--- a/AziAudio/ABI_Resample.cpp
+++ b/AziAudio/ABI_Resample.cpp
@@ -109,7 +109,7 @@ void RESAMPLE() {
 
 		// imul 
 
-		dst[MES(dstPtr)] = (s16)(pack_signed(IncrAccum(src, srcPtr, lut)));
+		dst[MES(dstPtr)] = pack_signed(IncrAccum(src, srcPtr, lut));
 		dstPtr++;
 		Accum += Pitch;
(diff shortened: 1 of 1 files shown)


cxd4:
Some good ideas here; the fourth commit should benefit future changes as well.

samuelyuan:
Thanks for your suggestion though. I feel that there are some areas in the code that have unnecessary casting; will look into other areas and try to fix it.

Azimer:
I am too busy with other things to regression test (it's actually quite time consuming *sigh*).  So I am going to merge and test later.  I can always fix it myself later. :)

Azimer:
This change checks out.  Thanks Samuel.

PR #89, opened 2015-05-26
Update VS2008 project

Frank-74:
Remove SafeABI from project.

### DIFF
--- a/AziAudio/AziAudio.vcproj
+++ b/AziAudio/AziAudio.vcproj
@@ -210,10 +210,6 @@
 					RelativePath=".\HLEMain.cpp"
 					>
 				</File>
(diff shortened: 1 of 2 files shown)


Frank-74:
I didn't disable SSE2 in the MinGW build which is needed to successfully build at the moment.

cxd4:
It builds for me with or without SSE2; I have not seen any difference in compile success with or without -msse2 or -DSSE2_SUPPORT.  I just had to make a minor, unrelated change to the prototypes for externally linked symbols, plus removing the deleted file that Azimer deleted like you did in this PR, which I'm sure @weinerschnitzel will want to do for the Xbox solution as well now.

PR #94, opened 2015-05-28
Fix 64-bit (and etc.) portability issues involving function pointer sizes.

cxd4:
The main reason I created this pull request (though there were other nitpicks of mine in the process) is because frequently we will find this in the HLE main source:
```c
memcpy(ABI, ABI1, 0x20 * 4);
```
What is wrong with this code?

Well it assumes two things:
* ABI[] and ABI1[] have 0x20 = 32 elements.
* The size of each element is 4 bytes:  `sizeof(ABI[0] == 4)`.

The first assumption isn't too bad since it's well-established as a matter of the test of time that only 32 high-level audio commands for HLE seem to exist (though I indulged myself by creating a macro for that anyway, despite my attempts to find documentation that 32 is the definitive limit), but the second assumption is fatal for 64-bit builds of this plugin.  We may not have any 64-bit zilmar spec for testing the plugin as of yet, but I can tell just by looking that this `memcpy` is not going to have the desired effect.

Any 64-bit implementation of the C language that I have heard of does not (and can not) have 4 as the size of a function pointer.
Correct should be either:
```c
memcpy(ABI, ABI1, 0x20 * sizeof(void *));
```
... or ...
```c
memcpy(ABI, ABI1, 0x20 * sizeof(p_func));
```
... where `p_func` is one of the types already defined in my `"my_types.h"` header, to avoid the rather ugly syntax of nested parenthesis involved with declaring tables of function pointers.  I defaulted to the latter out of personal preference but feel free to debate over anything decided here before a merge.


Azimer:
Yep nitpicking me again. :)  It seems fine to me, especially the 32bit vs 64bit function pointer.  Thanks!

PR #95, opened 2015-05-29
Initialize data in a loop

samuelyuan:
Few observations:
v[0]...v[15] gets data from the array mp3data via multiple xor operations (to be exact, it's 
in the form `v[i] = GetData(2 * (i ^ (i / 2) ^ (i / 4) ^ (i / 8)))`. 
v[16]....v[31] gets data in the form `v[i] = GetData(2 * i).

The way the values are set up, there is a lot of swapping among a pair of elements. I wrote out all the values and expanded

Just for a concrete example...
`v[0] = GetData(2 * (0 ^ 0 ^ 0 ^ 0)) = GetData(0x00)`
`v[1] = GetData(2 * (1 ^ 0 ^ 0 ^ 0)) = GetData(2 * 1) = GetData(0x02)`
`v[2] = GetData(2 * (2 ^ 1 ^ 0 ^ 0)) = GetData(2 * 3) = GetData(0x06)`
`v[3] = GetData(2 * (3 ^ 1 ^ 0 ^ 0)) = GetData(2 * 2) = GetData(0x04)`
etc...

I don't think there's any errors, but you could double check to make sure all the indicies are correct.

### DIFF
--- a/AziAudio/ABI3mp3.cpp
+++ b/AziAudio/ABI3mp3.cpp
@@ -283,25 +283,12 @@ void InnerLoop () {
 	// Part 1: 100% Accurate
 	int i;
 
-	v[ 0] = GetData(0x00); v[31] = GetData(0x3E); v[ 0] += v[31];
-	v[ 1] = GetData(0x02); v[30] = GetData(0x3C); v[ 1] += v[30];
-	v[ 2] = GetData(0x06); v[28] = GetData(0x38); v[ 2] += v[28];
-	v[ 3] = GetData(0x04); v[29] = GetData(0x3A); v[ 3] += v[29];
-
-	v[ 4] = GetData(0x0E); v[24] = GetData(0x30); v[ 4] += v[24];
-	v[ 5] = GetData(0x0C); v[25] = GetData(0x32); v[ 5] += v[25];
-	v[ 6] = GetData(0x08); v[27] = GetData(0x36); v[ 6] += v[27];
-	v[ 7] = GetData(0x0A); v[26] = GetData(0x34); v[ 7] += v[26];
-
-	v[ 8] = GetData(0x1E); v[16] = GetData(0x20); v[ 8] += v[16];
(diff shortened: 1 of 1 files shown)


cxd4:
It's fortunate that you observed a pattern there, though I'm tied on whether or not it will be beneficial to the finally optimized outcome.  It depends on how relevant the `(i / 1) ^ (i / 2) ^ (i / 4) ^ (i / 8)` pattern is to the intended mechanics of what the RSP microcode really intends...it's an unusual pattern so is difficult for me to say whether it will help vectorize the moves.

I do however like that you did this loop:
```c
for (i = 16; i < 32; i++)
    v[i] = GetData(2 * i);
```
Simple patterns like that are usually simple to vectorize.

Something as archaic and possibly uncertain on the other hand as:
```c
v[4] = GetData(0x0E);
v[5] = GetData(0x0C);
v[6] = GetData(0x08);
v[7] = GetData(0x0A);
```
would probably in the absence of the loop pattern require shuffling to vectorize the moves.

Something like
```c
s32 temp[32];
__m128i xmm;

for (i = 4; i < 8; i++)
    temp[i] = GetData(2 * i);
xmm = _mm_load_si128((__m128i *)&temp[0]);
xmm = _mm_shuffle_epi32(xmm, _MM_SHUFFLE(1, 0, 2, 3));
_mm_store_si128((__m128i *)&v[4], xmm);
```
However I'd say the vectorizing of all of what goes on within each of the `GetData` calls is much more important overall, so I'd say it's 50/50 either way with or without the current change.

samuelyuan:
I don't actually know if there was meant to be a pattern or not, but I noticed that all the indices somehow worked using the given formula. If there is indeed a pattern, I'm really hoping there is a way to simplify `(i / 1) ^ (i / 2) ^ (i / 4) ^ (i / 8)` because that looks way too long the way it is. Or there could be a completely different way to solve this, but the current method is what I could come up with for now.

Azimer:
We can always refactor again to allow for better vectorization.  This works for now though and compresses an unnecessarily long block without adding another function.

PR #99, opened 2015-05-31
Proper makefile and mingw fix

SwooshyCueb:
Here's a proper makefile for building with mingw. I've also fixed the problem of DllMain not being an external symbol in mingw builds and made the build system part of the build string.


SwooshyCueb:
I don't suppose it would be too difficult to tweak the batch script instead of outright getting rid of it, so I'll hammer on that in a bit.

As for the optimization, -O3 is used for everything unless we're compiling a debug build, which doesn't use any optimization switches. IIRC gcc/g++ defaults to -O2 in this case. Is there any reason we wouldn't want -O3 for everything?

cxd4:
-O3 is based on -O2 in that both of them desire the best performance possible, but the means and the extremes through which they shoot for that goal are significantly different.  -O3 is sort of like -O2 but with a lot of extra flags automatically enabled (such as forcefully in-lining larger functions and loops, implicit vectorization to SIMD, some previously bug-prone optimizations that can incur a cache miss) which can in several cases incur actually the opposite effect:  a performance penalty.  That penalty may not be visible, but it wouldn't be worth the significantly larger build size than the size of the binary after compiling with -O2 if -O3 is slower than -O2.  One of my plugins actually had horrible performance from -O3 compared to MSVC or -O2 builds due to the huge loss of instruction cache space because of the large RDP functions.

In essence, I like -O3 for compiling things that are problematic to performance, like emulators (even if it's HLE and already optimized), but I don't think -O3 is necessary for basic sources to things like, reading the N64 AI registers, configuring the audio state, initializing XAudio/DirectSound ... those things do not make a noticeable performance weigh to users.  So it would make sense to prefer -O2 or -Os in those cases for smaller binary size, which could even help free up instruction cache space for parts of the rest of the program.  It's a little bit of a balancing act.

>I don't suppose it would be too difficult to tweak the batch script instead of outright getting rid of it, so I'll hammer on that in a bit.

I would appreciate that actually yeah, if it doesn't seem inconvenient for you.  Maybe the batch script could be replaced by some variant of make file (GNU makefile, SGI makefile, etc.), but I normally prefer shell scripts for their neutrality in simple syntax.  I've tried to understand the make system but to this day doubt that I would be able to maintain necessary updates to one by myself.  (I also like being able to just double-cilck the batch script in an explorer window to build it.)

I speak mostly from inexperience on this, but as far as I know the only disadvantages to shell scripts are a) less connection to system environment for dynamically filling in some build configurations and b) not directly executable outside a command interpreter [mostly] resembling Unix-based (usually this means either Windows or maybe a few easy syntactical adjustments in rare cases).  So I tend to believe that the more neutral and smaller syntax kind of makes up for it--probably if the project gets big and complex enough Makefile would be the best thing for everyone, though in its current form it is just a single plugin.

SwooshyCueb:
Just added some new batch scripts. Nothing too fancy, really.

cxd4:
I guess moving -O2 to globally apply to all sources (rather than -Os, -O2, -O3 selectively for individual modules like I was doing with the original batch) isn't such a bad idea--probably for the better since it's less confusing to other people maintaining it.

Would it be possible to consolidate the 4 MinGW batch scripts you pushed into 1 script (just to keep most of the repository to the actual C sources)?  I think most people who just want the plugin aren't interested in testing the debug builds (or would be given instructions to change a batch variable to make it compile with debug info in it), so probably the two debug scripts could be removed.  As for 32- versus 64-bit builds, I guess having separate batch files for that makes the frequent testing of both easier, though I didn't think it was too hard to change `set MinGW=C:\MinGW\bin` to `set MinGW=C:\MinGW-w64\bin` either, as that's usually the only difference between 32- and 64-bit MinGW build scripts in my personal projects away from this one.

Other than that it looks better to me!  I never thought about building both the XA2 and DS8 plugins with 2 fast linker commands.

cxd4:
something like
```
set MinGW_32=C:\MinGW
set MinGW_64=C:\MinGW-w64\bin # MSYS users may prefer C:\msys64\mingw64

REM Set this to $MinGW_64 for 64-bit builds or $MinGW_32 for 32-bit builds.
set MinGW=$MinGW_32
```

SwooshyCueb:
Shouldn't be too difficult to merge them. I'll hammer on that later today.

AziAudio/common.h
```diff
@@ -64,14 +64,21 @@ unsigned long GenerateCRC (unsigned char *data, int size);
...
+#ifdef __GNUC__
+#define PLUGIN_BUILDSYS "Mingw"
+#else
+#define PLUGIN_BUILDSYS "MSVC"
+#endif
+
 #define PLUGIN_RELEASE " v0.70 "
 #define PLUGIN_BUILD "WIP 5" \
-	PLUGIN_DEBUG
+	" (" PLUGIN_DEBUG \
```
Azimer: I'm not sure we want this in the long run.  It will only confuse to have the plugin be "Azimer's XA2 Audio v0.70 WIP5 (MSVC)"  What is the relevancy of how it was compiled to the end user?  Alternatively, if this is for the sake of development, I would rather make this part of a #ifdef _DEVELOPER_BUILD_ or some other define to turn items like this on which are only relevant to developers during testing.  If you have a counter-argument I'd like to hear it.

Azimer:
Let me know when this is ready to merge and I will review it again.

SwooshyCueb:
As you suggested, @Azimer, I have changed the build string so that it only adds the build system if DEVBUILD is defined.

I've also merged the MSYS-less build scripts into one script. Should be ready for merge, unless there are any further objections

cxd4:
When I get back on Windows I'll be sure to double-check the new .cmd script, but it looks good to me.  My only potential issue with it was just that I liked to make build scripts as "portable" (as in, with the least amount of edits necessary) to other operating systems as possible.  This usually involved minimizing/avoiding all together the use of commands like ECHO, but the idea is this script is for building on Windows anyway without MSYS so this really shouldn't come across as an issue I don't think.  Most importantly your new script looks much easier for me to maintain in case of any future changes needed, as I can understand it much better now.

SwooshyCueb:
Actually, I just noticed a couple things that need fixing. Hold off on merging for now

SwooshyCueb:
Should be good to go now, @Azimer.

Azimer:
@cxd4 : I don't build with this environment.  Are you okay with everything you've seen here?

cxd4:
Since you ask, actually no, I'm not, but it compiles successfully.  The final commit of this pull request has complicated the build script into something power-fixed to the abilities of specific OS-tailoring enough for which I am not even sure I will be able to maintain in case of future updates to the source part of the repository.  However my advice would be don't let that stop you from merging this PR anyway; this repository exists because of the plugin source code, not various make scripts or makefiles.  So for now I would ignore that.  If it becomes an issue the next time I need to change something about the MinGW script to synchronize it to the audio sources, I will revert the new .cmd file myself and restore that to the way it was originally.

The complicated issue here is that there are different types of changes--we have to weigh the overall benefit to the overall "potentially" unnecessary things.  My personal preference would be go ahead to merge the PR, except that I am wondering about a couple of questions first.

I have no comments on the GNU Makefile.  Some modernists certainly prefer using them, but I think that source code is the relevant part to the repository rather than power extensions to dynamic ways to build a projects.  Therefore I don't think it matters--if anything the added Makefile should make up for the primitive batch file I had originally, rather than trying to do the same types of practical convenience and abilities in both files.

AziAudio/common.h
```diff
@@ -63,15 +63,28 @@ unsigned long GenerateCRC (unsigned char *data, int size);
...
+#ifdef __GNUC__
+#define PLUGIN_BUILDSYS "Mingw"
+#else
+#define PLUGIN_BUILDSYS "MSVC"
+#endif
+#ifdef _DEBUG
+#define PLUGIN_DEBUG " (" PLUGIN_BUILDSYS " Debug)"
+#else
+#define PLUGIN_DEBUG " (" PLUGIN_BUILDSYS ")"
+#endif
```
cxd4: By the way is this still necessary?  Why do we need to have this code added?

There are hundreds of compilers, all for legitimate C language implementations.  I like to think that C code should concentrate just on being conformant and compatible (well not to mention other obvious issues like performance) rather than printing compiler-specific stuff or extensions, unless helpful with something.  It's usually self-explanatory to me based on DLL file size alone which compiler was used to do the build (also easy to tell with utilities like dependency walker or even just hexing the compiled DLL to look for MinGW lib strings), and unless we swap alternatively back and forth between 2 compilers then even just our own memory by itself should be enough to tell us which compiler we just used to do the build.  (Some of the same probably could be said about printing "Debug" except that seems to have been Azimer's idea to begin with anyway.)

Azimer: I have a practical reason for _Debug. In my build environment everything goes to PJ 2.2 plugin folder and lets me know which plugin I am using since there is the possibility of having AziAudio.dll and AziAudio_d.dll.

cxd4:
I rarely test the Debug builds of any software myself so I guess that makes sense.  (I also rarely think to set build paths to coincide with $pj64/Plugin, which is another reason why it benefits you because then the same plugin name shows up twice instead of copying the dll through explorer like I usually do.)  zilmar was trying to get me into the habit of using the Visual Studio debugging tools by Microsoft but so far I've still been in business with the old sprintf/MessageBox/assert habits.

Azimer:
I use debug builds most often to test why post-conditions are not equaling the expected result given a set of pre-conditions for a given function or algorithm.  It's also why you saw a lot of __asm int 3;'s initially as they let me break quickly to step through.

I will go ahead and merge this PR then if you're okay with it.

cxd4:
For that, yeah I usually just sprintf() a bunch of variables and print them, and compare the values myself to induce where the expected result vs. actual result went wrong.  But yeah I'm sure there are some technical things that the Microsoft debugging makes possible to do, probably easier for some things.  Just not into the habit of it myself.

PR #101, opened 2015-06-03
OpenAL driver (don't merge yet)

SwooshyCueb:
I'm hoping this will be useful eventually.

It's pretty slow right now, and you have to close and reopen Project64 after ending emulation if you want to play again, but it works.

Granted, the only way I have of testing this at the moment is running it in wine.

I've also made separate VS projects for each version of the plugin (DS8, XA2, OpenAL) but the original VS project is still in there if we ever do merge them all into one plugin. It's just not in the solution.


SwooshyCueb:
Note to self: Look into context attribute settings on creation. See pg 100 of this PDF
http://web.archive.org/web/20130327063429/http://connect.creativelabs.com/openal/Documentation/OpenAL_Programmers_Guide.pdf

cxd4:
Whoops, commenting on the same sources from a different PR seems to have carried over here.

Well anyway, this I think would be awesome.  I was always thinking about doing an OpenAL implementation but lacked the particular knowledge on cycling individual sound pre-buffers--if we could have a simple, minimalist (avoid intereference with existing code) implementation of OpenAL into Azimer's plugin then we could finally start testing this plugin outside of Windows.  I would be happy if the points of this PR came to fruition.

Only issue is the size of all the changes...hangs up my browser if I try to analyze the whole diff, and many of the commits seem to be tied from the unrelated MinGW PR which has nothing to do with adding OpenAL code I think.

Azimer:
I highly recommend saving off your OpenAL stuff and redoing the PR with a clean base.

SwooshyCueb:
Will do.

SwooshyCueb:
There are a couple of other things I'd like to do anyway before this gets merged.

PR #102, opened 2015-06-10
internal fix to compiler truncation warnings:  uint16_t to int16_t

cxd4:
This keeps the warnings fixed when compiling with /W4 without the need for disabling the warning code using #pragma.  Maybe some compiler can still warn about storing int into int16_t without an explicit type cast, but if that is the case let me know and we can update the internal warning further.


Azimer:
I will test and verify once I return from my business trip. I just want to ensure it fixes the warning without causing a new problem.

cxd4:
Yep good idea.  And I thought about being "extra safe" by changing something more like this
```c
 // 0xfedf, 0x47ed, 0x3a95, 0xfeac,
    -0x0121, +0x47ED, +0x3A95, -0x0164
```
However IMO that takes long to manually squint everything to prove I didn't misconvert at least one of the values, so I would rather stick to just removing the explicit int16_t type casts at least for now.

PR #104, opened 2015-06-12
Implement device list

SwooshyCueb:
This probably needs some testing. I've done just about all the testing I can with wine and my development VM, but, frankly, that isn't a whole lot.

Anyway I've implemented the device list in the configuration dialog for both versions of the plugin.

(don't merge yet)


cxd4:
I can get around to testing it natively on Windows if it looks okay to Azimer.  Much with C++ I cannot understand so can't really comment about much otherwise.

SwooshyCueb:
It would help if you have more than one audio device and can make sure that switching works.

SwooshyCueb:
Actually, don't merge this just yet. I'm gonna see about adding a device list refresh button in the configuration dialog. I think I've also spotted a memory leak.

SwooshyCueb:
There, refresh button implemented and leak fixed.

Frank-74:
Just tested this and it doesn't change playback device.  I have the options M-Audio Delta 1010 1/2, M-Audio Delta 1010 3/4, M-Audio Delta 1010 5/6, M-Audio Delta 1010 7/8, M-Audio Delta 1010 Multi and M-Audio Delta 1010 S/PDIF.  But no matter which I select, playback always outputs from 1/2.

I have 'Use only default devices' unchecked in Windows Sounds and Audio Device properties.

SwooshyCueb:
Huh. Okay. Were you using the XA2 version or the DS8 version?

Frank-74:
Used the DS8 version.  The XA2 build failed from your branch using the mingw-build.cmd.

SwooshyCueb:
I'm not sure if that's caused by my changes or not, but I don't think it is, as I've had trouble with the XA2 build for a while now. Either way I'll see if I can't fix the device selector today. Thanks for testing

SwooshyCueb:
@Frank-74, before I start hammering on this, can you tell me if the XA2 version of the plugin fails from the master branch, too? I'd like to verify whether or not this is a problem specific to my device list branch.

Frank-74:
@SwooshyCueb Master branch builds both XA2 and DS8 with no errors.

SwooshyCueb:
Yeah I figured out it was my code, but in actually getting the XA2 version to choose a specific device I also fixed the problem it was having with wine, so I should be able to test XA2 builds as well as I can test the DS8 builds if we merge this.

Anyway, I think my branch is ready for testing again. Pretty sure the XA2 device switching works properly, but I'm not 100% sure about the DS8 version

Frank-74:
I still get the same error.

    XAudio2SoundDriver.cpp...
AziAudio/XAudio2SoundDriver.cpp:63:8: error: 'PROPERTYKEY' does not name a type
 static PROPERTYKEY PKEY_AudioEndpoint_Path = {

SwooshyCueb:
Looks like you don't have PROPERTYKEY defined in your wtypes.h. I've added a definition to the XA2 header. See if it builds now.

Frank-74:
@SwooshyCueb I can confirm that device choice does now work with DS8, though there are some problems.

I tested with Goldeneye.  Started Goldeneye, and selected M-Audio Delta 1010 5/6.  No sound.  End emulation and restarted Goldeneye, and I now have sound from M-Audio Delta 1010 5/6, but device selector says Primary Sound Driver.

I'll try rebuilding to see if it builds the XA2 version now.

Frank-74:
XA2 builds now.  Same problem as above, though don't have the Primary Sound Driver option with XA2.
After changing to 5/6 (nosound), ending emulation and restarting, selector shows 1/2, but output is from 5/6.

Azimer:
Is there any update on this PR?  I've been working on other non-plugin related projects waiting to see what happens with this PR before I make it not so easy to merge anymore.

SwooshyCueb:
I've been a little busy lately with my new job. Working retail for the first time and ow my feet. Should have plenty of time to hammer on this some on Friday though.

Azimer:
It's fine.  I've been very busy myself.  I don't want to lose track of this though since some of the work you've done would need to be done by me in the future.  It is entirely my goal to merge the backends to work on a common audio code and be interchangeable through the configuration.

cxd4:
Well, in the meantime there is a question I'm wondering (since I can't read too well into the API stuff being done in this PR).  How will the audio devices (DirectSound, OpenAL, XAudio2 possibly others to come?) be linked?

If the fruits of this PR get merged, would it mean that AziAudio.dll requires the runtime dependencies for all of those audio APIs just to be able to initialize the DLL to choose which one of them to use, or is there some dynamic voodoo being used here to not invoke any of those dependencies unless it can detect them or something?

I mostly ask just out of curiosity.  This is one of the fundamental debates over the plugin system (or at least the current zilmar plugin specs).  I suppose that Mupen64Plus attempts to fold this issue down by means of their own plugin system, but if we were to take something like libretro or RetroArch's implementation into perspective for challenges like this, then it would be kind of like having Project64 itself expose the different audio device APIs (DirectSound etc.) and the audio plugin having nothing to do with any audio device code, just the N64 RCP AI and RSP HLE.  (Of course we have many pros and cons for this as well.)

So it's a tricky subject, just made me kind of wonder what might come about if this PR succeeds.

Azimer:
Well... My thought process is to basically do all the AI Register, buffer emulation, and timing from the base class and allow each derived / API specific class implement the respective API and provide the interface for the virtual methods.  I also intend on allowing those class files to be commented out with preprocessor commands that will cause it to be excluded from both the configuration as selectable and remove the dependencies.

XAudio2 doesn't require a library for 2.7 but it is installed by default with the latest DirectX as well as DirectSound.  Anything else, I would hope everything is statically linked or it is a default install for Windows (or whichever OS it's installed for).  I hope not to bloat the size of the binary by providing a small handful of API options and allowing a design that would give us the opportunity to implement additional APIs as necessary.

Now with this PR, I am not sure where's it going yet.

cxd4:
Yeah commenting out/using the preprocessor to not link in most possible APIs was kind of the way I was thinking.  At least, it's what I know how to do anyway, and it saves file size.  It just didn't look like what this PR was doing--in this case it looks more like it's able to dynamically succeed in loading the plugin even if the APIs like OpenAL aren't installed.  Which is interesting, but yeah kind of heavy on size (so would be statically linking them).

Azimer:
I've started working on this again.  I plan to work on DirectSound and XAudio2 as a template then leave the rest to someone else to implement while I tackle more pressing issues.  I think integrating the device selection will also be necessary so that may end up causing this PR to be redundant.  I will see what I can use from the PR and give credit since I can't merge it as is anymore.

SwooshyCueb:
My apologies for letting this sit for so long. My new job is really sucking away all of my energy to do other things. Since I probably won't be able to work as consistently as I thought, I'll work on less important things, such as the OpenAL backend I proposed a while back. This way I won't hold up development.

Azimer:
No problem.  Keep the PR open until I close it.  I will reuse as much of your work as I can.

SwooshyCueb:
@Azimer, I just added a comment to one of the commits that might be helpful
https://github.com/SwooshyCueb/AziAudio/commit/5072ef0440a7faa9f5a888489764d72c21605294#commitcomment-12055194

cxd4:
Well, thanks to the No Sound driver that Azimer had last added, it seems more than plausible enough for me to tackle porting this plugin to Linux.  Maybe from there I can get a feel for the C++ class-oriented design with switching the APIs, and possibly implement an OpenAL sound driver option later (so that the Linux port actually has sound).  Another thing it makes me wonder is whether the hypothetical OpenAL sound driver (if I am able to add it) would replace the DirectSound one, since OpenAL is pretty much just a high-level API that, on Win32, runs on top of DirectSound (dsound 3D on Windows XP, software on Vista, dunno if they'll make it run on top of XAudio2 in the future).

Azimer:
I am hoping to have a handful of APIs implemented then decide which are the best by performance and implementation and have those be the default.  If that is OpenAL, it could very well replace DirectSound.  My goal is to keep it as dynamic as possible without forcing dependencies, for example, if you choose waveOut, you won't necessarily need DirectSound8 or XAudio2 installed in order for the plugin to run.  XAudio2 is easy since it uses COM.

cxd4:
Right...it probably should end up in my plan that OpenAL is tested on Windows, but macro'd out of the DLL linkage by default to not impose the dependency requirement.

Basically it would be the only implemented, cross-platform API for sound output (ALSA is Linux-only, XAudio is Windows-only, CoreAudio is Mac-only, OpenAL works on all of those without having to involve all the bloat of SDL), but most people don't have the OpenAL dependencies installed anyway so I wouldn't know how to enable it in Windows builds anyway without inconveniencing most people.

weinerschnitzel:
I'll be happy with moving the XBox port to a separate fork if OpenAL trumps DirectSound8 in all facets on the PC builds. I'd prefer if this change took place when things are finalized, compared and profiled, as Azimer suggests.

cxd4:
I forgot, on Xbox you would have to use DirectX, not OpenAL (since Xbox only has fully proprietary APIs available hehe :P).

Well...I suppose by "replace" DirectSound, I really should have said, keep the code around (or at the very least accessible), but consider commenting it out for official Win32 builds on desktop Windows since XAudio2 is better.

PR #108, opened 2015-07-08
64-bit compilation fix (macro preprocessor regression caused by me)

cxd4:
To begin, I would like to point out that this plugin does work on 64-bit Project64 2.x, _with some changes_.

However because zilmar currently is a Slowpoke (sorry :D) with merging the more important, needed changes so that the upstream and official Project64 64-bit works, I have made [64-bit Project64 1.4](https://github.com/cxd4/Project64_64) here, which also works with this plugin.

When I pulled from master I just had to fix a couple build issues, and it works again.

There was a crash with Zelda MM in 64-bit audio HLE but that seems to have been fixed by pulling the commits I pushed a while back to fix the function pointer size memcpy'ing in the HLEMain function.  So it looks like this plugin pretty much works flawlessly in 64-bit.

### DIFF
--- a/AziAudio/HLEMain.cpp
+++ b/AziAudio/HLEMain.cpp
@@ -352,7 +352,7 @@ void vsats64 (s16* vd, s32* vs)
     __m128i xmm;
 
     xmm = _mm_loadu_si128((__m128i *)vs);
-    xmm = _mm_packs_pi32(xmm, xmm);
+    xmm = _mm_packs_epi32(xmm, xmm);
     *(i64 *)vd = _mm_cvtsi128_si64(xmm);
 #elif defined(SSE1_SUPPORT) || defined(SSE2_SUPPORT)
     __m64 result, mmx_hi, mmx_lo;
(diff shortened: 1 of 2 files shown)


Azimer:
Glad 64bit is working!! That's awesome and good to know!

cxd4:
Yes, it took very surprisingly few changes.  I was sure that the interpreter op functions to the Project64 core would have ABI issues (long vs. int vs. DWORD) when compiling as 64-bit but that lucked out.

In fact with this audio plugin, the only thing broken with 64-bit was ABI2[] function pointer table:
https://github.com/Azimer/AziAudio/blob/master/AziAudio/ABI2.cpp#L24
... which used to cause crashing in audio HLE for games using ABI2, because ABI1 and ABI3 had a bunch of SPNOOP for the bottom half of the function table, while ABI2 had non-SPNOOP function pointers in the second half of this array.  After having already sent the PR to change the `memcpy` side from 32 * 4 to `NUM_ABI_COMMANDS * sizeof(p_func)` (where sizeof(p_func) would be 4 on 32-bit builds, 8 on 64-bit builds), seems the only problem this plugin had with 64-bit portability was already fixed.

PR #110, opened 2015-07-10
added option to build using the (older, XP-compatible) Windows DDK

cxd4:
I wanted to make an issue about this but needed to show how the script would look, so here the "issue" is in pull request form.  Possibly this should be rejected instead of merged; it depends what we want.

The reason I like to do this is so we can link against MSVCRT.DLL, instead of statically linking with /MT like zilmar does over at Project64 or dynamically linking MSVCR80, MSVCR90, ... whatever which people won't necessarily have installed on their system or understand why the plugin fails to initialize.

MSVCRT.DLL is the "known DLL" on all versions of Windows ever since Windows 95 b and is what the components to today's versions of the Windows operating system continue to use, despite forcing all other developers to use the MSVCR80, 90, 100... series of numbered and non-universal DLL's.  (The reason this started was because it became infeasible over time for Microsoft to update their C library over every release with the same unified "msvcrt.dll" filename with new features, such as MSVC or C99 extensions and support, which commonly serve no meaningful value for simple engineering.)

So maybe it sounds absurdly unorthodox to use the Windows Driver Kit and maybe these changes should be declined; I can also understand it that way if the CRT library issue takes overall less priority than the VS environment or debugging etc..  I had a couple good long reads from some of these people who also like to do this to centralize their standard C to a universally available, external library from these blogs:

https://kobyk.wordpress.com/2007/07/20/dynamically-linking-with-msvcrtdll-using-visual-c-2005/
http://www.syndicateofideas.com/posts/fighting-the-msvcrt-dll-hell

The result of these changes is that my AziAudio.dll is about 100 KB, while the AziAudio.dll produced from the VS2013 solution in this repository using /MT (static multi-threaded) is over 150 KB.  Screenshot of the new DLL's analysis in Dependency Walker:
azicrt

The only thing I have not figured out how to do is dynamically link a universal variant of the MSVCP* libraries (C++ counterparts to MSVCRT.DLL).  I guess in the WDK that gets statically linked, but there are only 4 C++ functions being used in AziAudio so maybe by removing those that issue could be subsetted out anyway down to just dynamically linked MSVCR, absolutely no MSVCP dependency (meaning neither static- nor dynamic-linked).

### DIFF
--- a/make_wdk.cmd
+++ b/make_wdk.cmd
@@ -1,0 +1,41 @@
+@ECHO OFF
+TITLE Windows Driver Kit 7.1.0
+
+set src=%CD%\AziAudio
+set obj=%CD%
+
+REM set target=i386
+set target=amd64
+
+set DDK=C:\WinDDK\7600.16385.1
+set MSVC=%DDK%\bin\x86\%target%
+set incl=/I"%DDK%\inc\crt" /I"%DDK%\inc\api" /I"%DDK%\inc\api\crt\stl60" /I"%src%\..\3rd Party\directx\include"
+set libs=/LIBPATH:"%DDK%\lib\crt\%target%" /LIBPATH:"%DDK%\lib\wnet\%target%"
+
+set C_FLAGS=/W4 /Ox /Ob2 /Gm /Zi /Oi /GS- /EHa /MD
+set LINK_FLAGS=%libs% kernel32.lib user32.lib ole32.lib /DLL
+
+set files=%src%\main.cpp^
+ %src%\ABI_Adpcm.cpp^
+ %src%\ABI_Buffers.cpp^
+ %src%\ABI_Envmixer.cpp^
+ %src%\ABI_Filters.cpp^
+ %src%\ABI_MixerInterleave.cpp^
+ %src%\ABI_Resample.cpp^
+ %src%\ABI1.cpp^
+ %src%\ABI2.cpp^
+ %src%\ABI3.cpp^
+ %src%\ABI3mp3.cpp^
+ %src%\DirectSoundDriver.cpp^
+ %src%\HLEMain.cpp^
+ %src%\SoundDriver.cpp^
+ %src%\WaveOut.cpp^
+ %src%\XAudio2SoundDriver.cpp^
+ %src%\Mupen64plusHLE\audio.c^
+ %src%\Mupen64plusHLE\memory.c^
+ %src%\Mupen64plusHLE\Mupen64Support.c^
+ %src%\Mupen64plusHLE\musyx.c
+
+%MSVC%\cl.exe %files% %incl% %C_FLAGS% /link /OUT:AziAudio.dll %LINK_FLAGS%
+
+pause
--- a/make_wdk.cmd
+++ b/make_wdk.cmd
@@ -13,7 +13,7 @@ set incl=/I"%DDK%\inc\crt" /I"%DDK%\inc\api" /I"%DDK%\inc\api\crt\stl60" /I"%src
 set libs=/LIBPATH:"%DDK%\lib\crt\%target%" /LIBPATH:"%DDK%\lib\wnet\%target%"
 
 set C_FLAGS=/W4 /Ox /Ob2 /Gm /Zi /Oi /GS- /EHa /MD
-set LINK_FLAGS=%libs% kernel32.lib user32.lib ole32.lib /DLL
+set LINK_FLAGS=%libs% kernel32.lib user32.lib ole32.lib %obj%\resource.res /DLL
 
 set files=%src%\main.cpp^
  %src%\ABI_Adpcm.cpp^
@@ -36,6 +36,7 @@ set files=%src%\main.cpp^
  %src%\Mupen64plusHLE\Mupen64Support.c^
  %src%\Mupen64plusHLE\musyx.c
 
+%DDK%\bin\x86\rc.exe %incl% /fo %obj%\resource.res %src%\resource.rc
 %MSVC%\cl.exe %files% %incl% %C_FLAGS% /link /OUT:AziAudio.dll %LINK_FLAGS%
 
 pause


Azimer:
Interesting thought process I'd have to think about.  The only area I am absolutely not certain about changing is the way the C++ classes are structured.  There is also the case where I may need to use exception handling for xaudio2 to overcome the crash that happens.  It may be something to consider once the plugin is done with active development.

cxd4:
Hm, did I accidentally change something about the C++ classes?  Or you meant something about the WDK that compiles them differently.  I know I was getting a warning from <iostream> in main.cpp until I added /EHa to the compile flags.

Right now the C++ stuff is statically linked inside AziAudio.dll--that's why the result of compiling with the WDK is 90 KB while the result of using VS2013 in this repo's current make system with /MD (dynamically link everything) is 65 KB.  It's kind of like MSVCP is statically linked, but it will only let me dynamically link MSVCR.  That's why I was thinking, well yeah either a different version of the WDK or maybe removing the C++ dependency altogether would fix that remaining difference in size.

Sintendo:
You're not actually supposed to link against MSVCRT.DLL, though.

http://blogs.msdn.com/b/oldnewthing/archive/2014/04/11/10516280.aspx

Azimer:
Thanks for linking to this article. I'd rather not link against msvcrt.dll considering the potential compatibility issues.

cxd4:
So then I will close this issue but keep the build script I pasted for future reference, as I will be using this to build AziAudio.

cxd4:
@Sintendo, we have been aware of that article for some time now but nothing on that page points out that MSVCRT.DLL should always be avoided.

That is a Microsoft article which discusses revisions and extensions to the CRT (such as C99 support and other modified behavior to a few standard library functions), but none of those are applicable to the specification-defined behavior of the C standard library functions.

For example, Microsoft is allowed to change what happens for a function like `strcpy` if the source and destination strings overlap (which is undefined behavior), but they are NOT allowed to change the behavior of the function otherwise without breaking defined conformance to the C language.

Furthermore, if you have ever used MinGW-GCC to build on Windows you may have noticed that those builds always use MSVCRT.DLL, linked dynamically.  So your reference to msvcrt.dll being better avoided isn't just under-cited; it's unrealistic in the face of adversity in all cases where it has been done successfully.  There are plenty of other things Microsoft doesn't "recommend" which turn out for the better; I have also linked to two articles here already which cite that this central deployment technique has been used successfully.

cxd4:
Also, paying attention to the basic type of examples given on that page you linked:

>For example, I remember there was a fix for a Y2K problem which caused one application to crash because the fix altered the stack usage in such a way that exposed an uninitialized variable bug.

Again, uninitialized variables don't warrant a caution for avoiding MSVCRT, because those also are undefined behavior.  Using the same central msvcrt DLL as the components of the operating system use doesn't create problems with code--it encourages noticing them and fixing them.

Azimer:
At a point we may want to revisit but until we finish all the enhancements we intend on making. We would risk wasting time if we hit one of those compatibility issues. If MinGW uses this how do you know it doesn't suffer across Windows versions? How does is C++ affected?

Azimer:
I suppose at that point... why not use MinGW?

cxd4:
The point I stated was that logically, it can't be that there are any direct compatibility issues.
Though, I would love to hear an example suggesting otherwise.

I suppose you could ask how I can prove a negative, but it's simply a matter of the language rules.

The purpose of the C run-time is to provide the C library functions (or C++ library functions), and the _correct_ definitions of those functions are controlled not by Microsoft, but by the standard and definition of the languages themselves.

As per the guarantees of the language itself, there is no legally valid way that Microsoft can fundamentally change any of the C89 standard library functions in msvcr or any CRT version of the DLL for that matter.  (And again, the examples on the article Sintendo linked were no such changes, but low-level, sub-implementation details that programmers should not be accessing in a portable application.)  To do so would, at that point, constitute a language which is not C or C++, but a custom language by Microsoft.

The biggest complaint I have ever heard from anyone using MSVCRT.DLL (e.g., on MinGW) was that printf/sprintf couldn't understand `"%z"`, which is a C99 extension to ANSI C and not a required feature of the library.  Properly and compatibly written C code **cannot have modified behavior in-between different versions of the MSVC run-time libraries, as a matter of the established definitions of the language itself**.

But your question about C++, I cannot answer because I am not a C++ programmer.  Actually I don't really see myself as a programmer at all much, more of a software engineer.

cxd4:
If a C function like `printf` changed radically between MSVCR120.DLL and MSVCR999.DLL, msvcrt.dll or whatever, never mind dynamically linking to the DLL causing a compatibility issue between different versions of Windows.  I'd be amazed if **statically** linking them didn't make the situation even worse--in fact on Linux that's exactly what we have to avoid to help make sure Linux binaries more likely run and function on different distros/versions.  The C library should be dynamically linked.

> If MinGW uses this how do you know it doesn't suffer across Windows versions?

MinGW builds do have some problems, but those are problems related more to thinkgs like lack of public documentation by MS, like winapi includes and libraries.  Of the complaints I have heard, nothing about the involvement of `msvcrt.dll` has ever been one of them.

What's more, the linker version for the required OS version is always 4.0--Windows 95 (unless building 64-bit).  I would think that any strict incompatibilities about the CRT between Windows 95 and now would have been known today if they weren't temporary ones--of which msvcrt.dll is not a temporary bug and would have caused them to be permanent.

So it's not the version of Windows that changes MinGW builds as far as the topic of this pull request goes--it's the version of the C run-time being linked e.g. msvcr80, msvcr90, which varies less in terms of operating system, more in terms of the version of MSVC installed.

>I suppose at that point... why not use MinGW? 

For a long time, MinGW was my only option for this.  I liked it for cases like the RSP plugin where it was required for fully optimizing and I was able to refer away most of the C run-time functions/dependencies, but linking -lmsvcrt via MinGW typically links in some small sections of extra sectional crap and startup code.  A simple "Hello world" application built by MinGW to use msvcrt.dll would be today around 8 KB, while on MSVC with /MD it would be no more than 2 KB.

That is why I do not always prefer the MinGW approach on Windows, but you're right.

>At a point we may want to revisit but until we finish all the enhancements we intend on making. We would risk wasting time if we hit one of those compatibility issues.

If and when the situation arises where time has been wasted by one of these "potential issues", the only one risking the waste of that time is Microsoft for breaking language conformance rules and turning C into pseudo-C or some other language.  So far I have never seen any evidence of this happening with MSVCR.

Either way you're right that it takes no more priority over the other enhancements and goals to this plugin, and I was prepared to close this issue anyway and expected the idea to be seen as unorthodox.  Even so I was genuinely curious what everyone's thoughts on this would be.

Sintendo:
I want to clarify I wasn't necessarily objecting to this change. I felt that Microsoft's stance on linking against MSVCRT.DLL was relevant and posted the link. I agree that linking against it is unlikely to cause serious problems for most use-cases.

Azimer:
Understood. I still would rather revisit when I am done making source file changes. That should be in august. My free time is about to open up again.

cxd4:
Hm.  Well GitHub seems to be showing a "Restore Branch" feature but the last issue with this pull request I wanted to retry was that the C++ stuff is still statically linked.  (There's a MSVCRT.DLL but not a MSVCPT or whatever.)  Just made me kind of feel weird dynamically linking MSVCRT but still statically linking C++ stuff.  Hopefully the classes don't have anything to do with it and it's just stuff like `<iostream>` that's causing it which could be removed or replaced; otherwise all I can try is to install a different version of the WDK and hope that changes.  There is about a 30 KB difference from static-linking the C++ run-time alone.

Microsoft's stance most likely would be relevant, against using msvcrt, if it wasn't incomplete representation every time using scare-tactic examples.  They themselves have _been_ using msvcrt in the operating system DLLs...shouldn't they be linking their OS components against msvcr80.dll on Windows XP, msvcr90 on Vista, etc. or whatever?  But they don't.  And they even provide a version of MSVC that uses msvcrt just like they do.  msvcrt.dll has probably the most restrictions in terms of _optional, extra_ features that the C language doesn't offer or guarantee, but that only helps portability, not detriments from it.  And somehow I don't think MS has had a history of caring about that sort of thing.  Not like proper engineering is any better than forcing everyone to install 1000 copies of the same msvcr### all over their computer every windows update or anything.

PR #113, opened 2015-07-22
Fixes/Updates for XBox port

weinerschnitzel:
Small changes to keep the XBox port up to date, but let me know if "Gitiquette" would prefer me to do smaller PR's. I think it's ok, but I'm also very unorganized. :P


cxd4:
Seems just fine to me.

cxd4:
Actually this reminds me,
```c
#pragma warning(disable : 4100)
#include <dsound.h>
#pragma warning(default : 4100)
```
Hate disabling warnings that way, and it's usually compiler-specific anyway and isn't significant to C code itself.  I think after this PR is merged (just so I avoid merge conflicts) I will move that out to common.h as well.

weinerschnitzel:
Agreed. Sounds good to me. :)

Azimer:
Necessary evil having warnings set the way they are.  I am really thinking of changing the property sheets to remove that annoyance.  It's either that or changing the dsound.h include.

Does dsound.h belong in common.h though?

cxd4:
No, it doesn't really belong in `common.h`.  I wasn't really thinking about that issue since I was more focused on saying #pragma warning disable in more than one file...if dsound.h only needs to be included by a single maybe two .cpp files then I guess it's not really an issue to care to move it to common.h.

cxd4:
I'm happy with using warning level /W4 like zilmar kept it.  The only thing I really dislike about the property sheets is statically linking the CRT every single time.

weinerschnitzel:
Tried to add those to another branch, sorry.

weinerschnitzel:
Can I finish? Can I finish? OK, I'm finished. Plugin is linking with Surreal64 with ease. I'll have a PR up for vs2003 project files, although I'm probably the only one still with vs2003 at this point...

cxd4:
yea I would tend to disagree with the point to a vs2003 solution but don't want to confuse discussions here with one that, hasn't been created yet, so yeah I can bark questions about that once the issue/PR is started :P 

But everything in this PR is straightforward to me it seems.  The changes to dsound.h or removing the #pragma warning disable/centralizing it to common.h/etc. is just something I'm wondering for a future PR I might do.

Azimer:
As far as a 2013 solution I feel we are trying to support too many compilers already when a specific fork kept in sync might be a better option. I will merge this right away.

weinerschnitzel:
Understood, that's why I wanted to keep it out of this PR. Guess I'll have to get myself a newer compiler for windows builds. :p 

@cxd4, as for VC7 defining _WIN32, I'll take a closer look and see if another Surreal64 component is defining _WIN32. The issue became apparent when linking. I'll correct the commit when I resolve it on my end

cxd4:
I always wanted VS2003 so I could compile with /ML, but I never found it.  I'm no longer that interested in it since I can use the WDK to link against msvcrt.dll, without having to go through MinGW.

Instead of VS2003, I really think that 2005 sets the best minimum threshold, for three reasons:
* VS2005 is the first version of Visual Studio that introduces Express editions, so you can download them without having to pay money.
* It's the first version of VS to have an available 64-bit compiler and linker.  VC7 or VC6 are 32-bit only and are only useful to link against the 32-bit msvcrt.dll but not the 64-bit one.
* It's also the first version to use the "VCBuild" system, with .vcproj files.  Everything before that including VC7 used .dsw/.dsp files I think, so VS2008 can load 2005 solutions and vice versa even I believe (with maybe the change of a single line), up until VS2010 anyway.
* The numbered C run-time that MSVC 2005 uses is `MSVCR80.DLL` which is the first-ever numbered crt dll to have a 64-bit version to it.  So instead of linking against msvcrt.dll, another possible way to minimize arbitration is to just use the oldest numbered crt dll that is available in 64-bit.

>I'll take a closer look and see if another Surreal64 component is defining _WIN32.

Yeah that would most likely have to be a bug, since _WIN32 is usually a built-in define internal to the compiler.  It's not even a default to the preprocessor definitions in a solution file to VS or whatever; it's just permanently etched into MSVC.  So yeah that shouldn't be getting defined, as that is a conflict.

cxd4:
By the way I don't really care. :D

I just posted all of that because I felt like letting loose, not because I think we need any more .sln files in the repo.  Mostly just comparing between 2003 and 2005.

PR #114, opened 2015-07-23
added option to build using XP64-compatible WDK

cxd4:
https://github.com/Azimer/AziAudio/pull/110

Because Azimer mentioned wanting to investigate these changes in the future when he had time, I needed to push another commit to this (deleted) branch to fix the missing GUI resource script I neglected to test.  I'm merely recreating this branch in the form of a fresh pull request to avoid that confusion in case this does get investigated (and because deleting and recovering branches on GitHub didn't work the way I expected...).

### DIFF
--- a/make_wdk.cmd
+++ b/make_wdk.cmd
@@ -1,0 +1,41 @@
+@ECHO OFF
+TITLE Windows Driver Kit 7.1.0
+
+set src=%CD%\AziAudio
+set obj=%CD%
+
+REM set target=i386
+set target=amd64
+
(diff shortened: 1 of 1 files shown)


Azimer:
This is just one cmd file so I will commit it but continue to use vs2013.

cxd4:
Yeah.  Not trying to pollute the repo workspace with even further build files sorry; just aiming to make portability tests across targets/versions of Windows more convenient.  Feel free to move them wherever along the way.

PR #116, opened 2015-07-23
fixed warning (and probably bug) about assignment in condition

cxd4:
I am not able to tell by testing what bug (if any) this fixes, but it looks like a bug.


Azimer:
Doesn't fix a big that was important since the volume slider was only partially implemented.  This commit though conflicts with your previous one and can't merge automatically. So unless you resolve those it will have to wait until I can get some computer time or I get a full version of windows on my phone.  :)

cxd4:
lol??  It was a surprise this time.  I didn't expect it to be a merge conflict.

Git is complaining that the line **ABOVE** the line I changed in this single-line PR was changed.  It's a merge conflict because the end-of-line marker of the previous line, which I did not touch, connects to the start of the line that I did touch.

How absurd lmao.  will fix shortly

cxd4:
fixed and yeah, bug fix here is too insignificant to notice by running the plugin. :)

Issue #121, opened 2015-08-16
question about ADDMIXER from ABI2


cxd4:
This function looks null ever since before the source code to the old 0.55.1 versions of this plugin, in which the function still looked like this:
```c
void ADDMIXER () {
	short Count   = (k0 >> 12) & 0x00ff0;
	u16 InBuffer  = (t9 >> 16);
	u16 OutBuffer = t9 & 0xffff;

	s16 *inp, *outp;
	s32 temp;
	inp  = (s16 *)(BufferSpace + InBuffer);
	outp = (s16 *)(BufferSpace + OutBuffer);
	for (int cntr = 0; cntr < Count; cntr+=2) {
		temp = *outp + *inp;
		if (temp > 32767)  temp = 32767; if (temp < -32768) temp = -32768;
		outp++;	inp++;
	}
}
```

The same function can be seen in the current sources at:
https://github.com/Azimer/AziAudio/blob/master/AziAudio/ABI_MixerInterleave.cpp#L14

My question is:  What does this function accomplish?

Because for both old and new sources to this plugin, it seems to be a null function which writes back nothing.  Maybe that `for` loop is missing a statement to write the results of `temp` back to the buffer space.

Azimer:
Looks like a bug.  It should be *outp = (u16)temp; after the ifs.

cxd4:
Right.  I will close this as soon as I can find and prove where in a game this merged change improves the audio...I believe endianness doesn't matter in the commit I made (due to free boundary optimizations) but just want to link this to some game sounding better before I close.

cxd4:
Took me a good while to even find a game that quickly uses ADDMIXER.

`StarFox 64` immediately depends on ADDMIXER being implemented, soon after booting, but I can't tell a difference in the audio before or after this fix or with LLE/HLE audio.  So I guess it's safe to say...I didn't create a regression, though it's hard to tell by hearing if anything's better either.

Azimer:
That's likely why I never noticed it.  I'm not sure under what context it gets used in the API.

PR #122, opened 2015-08-17
ADDMIXER:  fixed null function by adding writeback

cxd4:
Seems to have been a bug since before AziAudio 0.55.1 old versions, as discussed here:  https://github.com/Azimer/AziAudio/issues/121

I don't know yet what parts of audio in what games this change fixes.


cxd4:
Due to endianness issues (should it be BufferSpace offset ^ 3??) it may be best to find what this improves.  If I do I'll post again.

Issue #129, opened 2015-10-05
How should we clean the repository root files up?


cxd4:
Now that AziAudio compiles and runs on Linux (no sound), I start to want to upload a new shell script called `make.sh` (also builds on Windows too if you execute it through Git/MSYS shell).

However as Azimer noted we have quite a plentiful lot of build script/project files in this repository already, and I am starting to want to move some of them somewhere.

The VS2013 solution is the only one Azimer uses IIRC and is recommended for Windows builds.

Maybe we could clean the root workspace of the repo by moving these files:
* make_wdk.cmd (my WinXP-64-bit-compatible build script to use MSVCRT.DLL)
* mingw-build.cmd (SwooshyCareb's modern, more complex MinGW build script)
* make.sh (not yet uploaded--my Unix shell script to compile on *nix-based systems)
* possibly the Xbox solution @weinerschnitzel contributed??

The fastest way to move all of these out of the root index of the repo, without creating a new directory, would be to put them under "3rd Party" I guess.

I can take care of this for the first 3 files (don't have Xbox/MSVC 2002 so can't test the 4th) by myself but would like anyone's opinion or at least @Azimer's on where this stuff could be moved to?

PropertySheets...zilmar is starting to wonder what to do with those in case he starts to think that separate repositories can help port Project64 easier.  (They may have to be changed/deleted.)

weinerschnitzel:
Maybe instead of moving the project files and build scripts to "3rd Party," how about we put them in a "Make" folder, and use another folder "Platform" to hold API specific headers/libraries and Xbox stubs?

cxd4:
Yeah "Make" sounds fine.

In the short term I wondered otherwise because atm we'd be moving 2 or 3 files and creating 1 folder.

>and use another folder "Platform" to hold API specific headers/libraries and Xbox stubs?

Well if we did that wouldn't this contradict the existing "3rd Party" folder which zilmar added?

weinerschnitzel:
It would be the same folder, I just feel that renaming it to "Platform" would make more sense.

cxd4:
I would have named it to something different than "3rd Party" too (at least without spaces), but this was the directory that zilmar chose and it was easier for us to match what was forward-compatible with what he was going to fork.  (If we do change that, it will collide with the rest of the PJ64 plugin/projects' setups.)

[...]

PR #134, opened 2015-10-06
Added Unix shell script for building outside Windows.

cxd4:
This new file I've created, `make.sh`, will compile the plugin either on Unix or Linux-based platforms or also on Windows, if you run the script through Git Bash/MSYS terminal to emulate the Unix command shell on Windows.  (If you do this, it compiles on Windows using MinGW, or I suppose Cygwin in theory.  The primary point however is to have a Linux build script.)

In the process, I've attempted simplifications of several build files we had formerly in the repo root.
See issue https://github.com/Azimer/AziAudio/issues/129 for more discussion.


weinerschnitzel:
Next step is to move the solution files to the Build directory?

cxd4:
Um, potentially, though I am happy to leave that to other contributors.

I'm no good at the MSVC project syntax/solution files and would need to have Xbox dev skills to move your Xbox solution into the build folder without fear of breaking anything.

I assumed people might prefer leaving the VC2013 solution as-is in the root directory since that's the official build method Azimer and most use.

weinerschnitzel:
Organization tells me to put them all in Build, or just leave them together. Of course, if Azimer just wants the vs2013 solution at root, that's fine as well. I'll keep the Xbox solution building like I promised, I just want to be on the same page of understanding if we are moving these solutions or not.

Frankly I'm also happy leaving them where they are. :) Whatever's clever.

cxd4:
My personal preference would be to see the Xbox solution moved under the "Build" folder.

I, individually, would not exactly mind if the same were done with VC2008 and VC2013 files as well, but I prefer to leave that for Azimer to decide.  Part of the issue could be how zilmar has the property sheets integrating with his 2008 solution and the shared folder with 2013.

Technically, since it is Azimer's repository and not my repository, my view is that my opinions can be moved to the "Build" subfolder and Azimer's opinions can be kept to the root folder.  Sure, if this was my repo I may have done it the other way around (e.g., my opinion is that `make.sh` is the most important build file), but all I was really concerned with was just cleaning up the root organization.  I understand that it's a little inconsistent to only move most, but not all, of the build files to the same folder, but I think that in exchange it's cleaner than it was before.

PR #136, opened 2015-10-23
Fix & Reorganize XBox solution

weinerschnitzel:
Performing the source code clean up as requested. Also fixed some build errors in the XBox project. Keep in mind that the MS XDK defines XBOX projects as WIN32 projects, only crippled. "#if defined(_WIN32)" macros must be coupled with "&& !defined(_XBOX)" in cases where WIN32 code is unsupported. Mostly GUI related.


cxd4:
Everything else me like!

weinerschnitzel:
Yeah... There is an automatically included Assert.h whose only purpose is to check if _WIN32 is not defined. Even though there is no code defining _WIN32 or any preprocessor definitions within the project file, the XDK automatically defines it. If I #undef _WIN32, the build fails. Also, AziAudio for XBox is setup to build an independent static .lib, that later links with Surreal's emulators. Sadly, explicitly checking if _XBOX is not defined will be a necessary evil.

cxd4:
I understand then.  Seems I will have to think of _WIN32 in terms of Xbox as well as desktop Windows from now on, whenever I write code in my other projects to be agnostic.  Thanks for the info.

weinerschnitzel:
Thanks for advocating platform agnosticism. :) I'll be be able to correct any issues as I see them here, the only thing to be attentive to is GUI code.

cxd4:
Getting too swamped by Project64 notifications to see this PR, @Azimer ?

Or just too busy with life to comment on a questionable issue with this PR?

weinerschnitzel:
It's the holiday season... I'm sure work family and life is taking precedence.

cxd4:
Oh no doubts about that, but you filed this Oct 22 over 30 days ago.  I just assumed he attempted to check his GitHub notifications at least once during that period but got swamped by other notifications.

Azimer:
Eighty hour work weeks and this is the first time I got to log into github and see a notification for it.

cxd4:
T_T

PR #138, opened 2015-12-01
Attach NoSound driver for non-Windows, non-Xbox builds.

cxd4:
Alright, my previous approach at a Linux build was a bit short-sighted.

While it does compile on Linux, I've underestimated the degree to which the AI and HLE functions all depend on the `snd` object to store a valid pointer to allocated sound driver class.  Simply keeping it NULL appears to not be an option that goes with the flow of Azimer's intended design.

So, I'm making this little change to attach NoSound driver #ifndef _WIN32 (includes _XBOX) as well.

Yes it will break compiling on Linux with a couple errors about WIN32-isms in the NoSound driver class, but I will be fine to tackle those based on whether these changes look acceptable.

### DIFF
--- a/AziAudio/main.cpp
+++ b/AziAudio/main.cpp
@@ -136,14 +136,12 @@ EXPORT Boolean CALL InitiateAudio(AUDIO_INFO Audio_Info) {
 /*
  * To do:  We currently have no sound-playing device for Unix-based platforms.
  */
-#if defined(_WIN32) || defined(_XBOX)
-#ifndef LEGACY_SOUND_DRIVER
+#if !defined(LEGACY_SOUND_DRIVER) || !defined(_WIN32)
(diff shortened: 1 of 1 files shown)


Azimer:
The end goal will be a user selectable option rather than a compile-time option with options being determined at compile time.  nosound/dsound/xaudio2 for Windows and nosound/SDL or something for Linux.

cxd4:
I understand the goal.  I haven't forgotten it but just haven't prioritized it.  I prefer the compile-time simplicity personally but understand that for Windows this has a higher advantage to implement.

After I concentrate on getting audio to work at all first, then I'll see about having the Linux build be configurable.  SDL audio can run on top of a number of lower-level audio APIs which makes this easy to switch.

PR #141, opened 2015-12-18
Fix regression in AiReadLength

LegendOfDragoon:
audioIsInitialized always = false, when I checked. This made it always
return 0. It already checks if audio is initialized in
snd->AI_ReadLength(). This commit fixes the problem in WDC where there
is no sound when fixed audio timing is disabled.

### DIFF
--- a/AziAudio/main.cpp
+++ b/AziAudio/main.cpp
@@ -254,7 +254,7 @@ EXPORT void CALL AiLenChanged(void)
 EXPORT u32 CALL AiReadLength(void) {
 	if (snd == NULL)
 		return 0;
-	if (audioIsInitialized == FALSE) return 0;
+
 	*AudioInfo.AI_LEN_REG = snd->AI_ReadLength();
 	return *AudioInfo.AI_LEN_REG;
 


cxd4:
Sounds like a good fix then, but in this case wouldn't it be preferential to cause `audioIsInitialized` to be set to `true` for WDC before AiReadLength gets called?  Just to be able to keep the current check in.

LegendOfDragoon:
What confused me is that in SoundDriver.cpp, it uses m_audioIsInitialized. So I'm not sure what audioIsInitialized should be used for.

Azimer:
Perhaps 0 is not the best value to return by default.  I am guessing returning *AudioInfo.AI_LEN_REG is more appropriate and fix your issue.  Please give this a try.  What we do not want to have happen is to cause a crash from calling a SoundDriver with an uninitialized state.

LegendOfDragoon:
Sorry, was mostly focusing on the DS8 code. I didn't realize that the XAudio2 code didn't do the same type of checking.

Basically, with LEGACY_SOUND_DRIVER, it checks if it's initialized here https://github.com/Azimer/AziAudio/blob/master/AziAudio/SoundDriver.cpp#L88. The way you had it set up, m_audioIsInitialized gets set in AI_Startup, while the variable audioIsInitialized never gets set anywhere. So I assumed you wanted to replace audioIsInitialized with m_audioIsInitialized, but didn't notice that you only use it for the LEGACY_SOUND_DRIVER.

In SoundDriver::AI_LenChanged, how come you only check if audio is initialized in LEGACY_SOUND_DRIVER? Just curious, so that I can understand your code better :smile: .

Azimer:
LEGACY_SOUND_DRIVER blocked code will be deleted once I put the time into converting the DirectSound and XAudio2 code to use the new SoundDriver superclass.

Azimer:
@LegendOfDragoon - Could you please verify your issue is resolve with my latest commit referencing this PR?

LegendOfDragoon:
@Azimer I still get no sound in WDC with FAT off. In your code, audioIsInitialized is never set to true, anywhere.

Azimer:
I understand where you're coming from now. I will look into it more. Thanks for clarifying.

Azimer:
Closing PR.  7affd42b860f1978b64d56072452abe9eab32ce9 is effectively this PR now.  Thanks for your contribution @LegendOfDragoon.

PR #142, opened 2015-12-22
Finally get the plugin to actually run on Linux.

cxd4:
Okay, it took me a few days to get over the mental blocks involved, but I've decided on a temporary way to modify the new, anti-LEGACY_SOUND_DRIVER code in `SoundDriver.cpp` to be able to compile on Linux.  Most of the stuff being macro'd out there is about the Windows mutex object and the "wait for object" functions, for which I need to extract a suitable replacement in terms of SDL.

Not that I am any fan of SDL, but the fundamental design of this plugin's newer AI functions to use advanced timing and object functions would overwhelm me if I didn't at least begin to include it, for now.

Since I've already decided to attempt replacing the newer SoundDriver Win32 class code with SDL-ism's, I figured it seemed opportune to also start working on a SDLSoundDriver.cpp -- I have obviously not finished it yet but included a new un-attached header, `SDLSoundDriver.h`, to show my current prototype for beginning it.

Emulation starts on Linux now, but the screen is just black with no output.  I'm not sure if that's the way that AI interrupts are expected to work under Mupen64 or if it's because of my unfinished SDL port of the driver code, but I'll find out soon enough.

### DIFF
--- a/AziAudio/NoSoundDriver.cpp
+++ b/AziAudio/NoSoundDriver.cpp
@@ -35,11 +35,16 @@ void NoSoundDriver::AiUpdate(Boolean Wait)
 #if defined(_WIN32) || defined(_XBOX)
 	if (Wait == TRUE)
 		Sleep(1);
+#else
+	if (Wait == TRUE)
+		SDL_Delay(1);
 #endif
 	if (isPlaying == true && countsPerSample.QuadPart > 0)
 	{
 #ifdef _WIN32
 		QueryPerformanceCounter(&perfTimer);
+#else
+		// To do:  Replace this with SDL, or Linux audio won't play.
 #endif
 		sampleInterval.QuadPart = perfTimer.QuadPart - perfLast.QuadPart;
 		samples = (long)(sampleInterval.QuadPart / countsPerSample.QuadPart);
@@ -75,5 +80,7 @@ void NoSoundDriver::SetFrequency(u32 Frequency)
 	countsPerSample.QuadPart = perfFreq.QuadPart / SamplesPerSecond;
 	QueryPerformanceCounter(&perfTimer);
 	perfLast.QuadPart = perfTimer.QuadPart;
+#else
+	// To do:  Replace this with SDL, or Linux audio won't play.
 #endif
 }
--- a/AziAudio/SDLSoundDriver.h
+++ b/AziAudio/SDLSoundDriver.h
@@ -1,0 +1,38 @@
+/****************************************************************************
+*                                                                           *
+* Azimer's HLE Audio Plugin for Project64 Compatible N64 Emulators          *
+* http://www.apollo64.com/                                                  *
+* Copyright (C) 2000-2015 Azimer. All rights reserved.                      *
+*                                                                           *
+* License:                                                                  *
+* GNU/GPLv2 http://www.gnu.org/licenses/gpl-2.0.html                        *
+*                                                                           *
+****************************************************************************/
+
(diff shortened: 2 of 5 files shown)


cxd4:
Looks like it's the former.

I guess it's been a long while since we've tested this plugin on non-Project64 emulators.  Something about the way Mupen64 vs. the way Project64 assumes interrupts to be done is not working with this plugin, even on Windows (but with the _WIN32 stuff all disabled).  I'm just glad it's not caused or introduced by my portability changes and will look into that later.

Azimer:
The old Mupen64 (not the plus variant)?  I can take a look at some point.  I will also take a look at your work today to and see if I can merge it today yet.

cxd4:
I cannot really test Mupen64Plus due to their direct incompatibility with any plugins.

Sorry though, I spoke too soon.  I assumed it wasn't caused by my portability changes because compiling the same code for Mupen64 0.5 for Windows had the same black screen as on Linux, except on Project64 2.2 it worked fine.  Looks like it's a "combination" of my temporary portability changes as well as the way Mupen64 handles window events from your AI code that causes the issue on Linux.

cxd4:
After further research, here is a list of _WIN32 blocks that, when disabled, causes Mupen64 0.5 for Windows to fail with this plugin in the same way that Mupen64 0.5 for Linux fails:

1.  https://github.com/Azimer/AziAudio/blob/07a6f255fdb08665aab41a4be0a92a00fb19ffc7/AziAudio/NoSoundDriver.cpp#L35-L37
2.  https://github.com/Azimer/AziAudio/blob/07a6f255fdb08665aab41a4be0a92a00fb19ffc7/AziAudio/NoSoundDriver.cpp#L41-L43
3.  https://github.com/Azimer/AziAudio/blob/07a6f255fdb08665aab41a4be0a92a00fb19ffc7/AziAudio/NoSoundDriver.cpp#L71-L78

The first of those 3, I can fix right now with this new commit I've just pushed to this PR:  https://github.com/cxd4/AziAudio/commit/43ba6946c1dd915c8864d15fd24f4c3f2081f1f1
The other two I will have to look up how to use SDL to replace later on.

Thankfully, nothing about how I touched `SoundDriver.cpp` seems to have any effect on Mupen64.

Azimer:
I like what you did but I think it should be refactored to be cleaner.  I am seeing this project with ifdef's littered everywhere and it shouldn't be that way.  Let me put some thought to it... then merge your code and refactor later.

cxd4:
There are some ways to get around that.

1.  My favorite:  Don't use any Win32- or Linux-specific code.  This is by far the cleanest solution.
2.  Keep the #ifdef _WIN32 away from the .CPP files, and define functions from the .H header files.  Then just call those (possibly macro) functions inside the .CPP files without any #ifdef.
3.  Probably some weird C++ language constructs that allow centralizing away most of the #ifdef.

Now option number 1 may seem impossible at times for your project, but it's not really that inconceivable.  For example, if AziAudio used an INI config file or whatever, then you wouldn't really **need** to #ifdef out the _WIN32 ConfigDlgProc/GUI for configuration--you could just delete all of it entirely for all builds with no macro, then people could use the INI file on Windows to change settings.

May sound rather drastic, but that's just for the purposes of example.  #ifdef everywhere can be avoided, but it requires careful planning and initiative to consider all means of avoiding using code from complicated APIs like windows.h which totally litter the compilation even more than the #ifdef macros themselves do anyway.

PR #156, opened 2017-02-07
Cheap Save/Load config settings

Frank-74:
I kept the same default settings.  Saves and loads sync settings, AI Emulation setting, and volume level, to Config\AziCfg.bin.

### DIFF
--- a/AziAudio/main.cpp
+++ b/AziAudio/main.cpp
@@ -174,7 +174,7 @@ EXPORT Boolean CALL InitiateAudio(AUDIO_INFO Audio_Info) {
 	file = fopen("Config/AziCfg.bin", "ab+");
 	fseek(file, 0, SEEK_END);
 	size = ftell(file);
-	if (size != 4)
+	if (size == 0)
 	{
 		fprintf(file, "%c", 1); // snd->configSyncAudio
 		fprintf(file, "%c", 0); // snd->configForceSync
--- a/AziAudio/main.cpp
+++ b/AziAudio/main.cpp
@@ -171,27 +171,30 @@ EXPORT Boolean CALL InitiateAudio(AUDIO_INFO Audio_Info) {
 	int size;
 	unsigned char *azicfg;
 	FILE *file;
-	file = fopen("Config/AziCfg.bin", "ab+");
-	fseek(file, 0, SEEK_END);
-	size = ftell(file);
-	if (size == 0)
+	file = fopen("Config/AziCfg.bin", "rb");
+	if (file == NULL)
 	{
-		fprintf(file, "%c", 1); // snd->configSyncAudio
-		fprintf(file, "%c", 0); // snd->configForceSync
-		fprintf(file, "%c", 1); // snd->configAIEmulation
-		fprintf(file, "%c", 0); // snd->configVolume
+		snd->configSyncAudio = true;
+		snd->configForceSync = false;
+		snd->configAIEmulation = true;
+		snd->configVolume = 0;
 	}
-	fseek(file, 0, SEEK_SET);
-	azicfg = (unsigned char*)malloc(size);
-	fread(azicfg, size, 1, file);
-	fclose(file);
+	else
+	{
+		fseek(file, 0, SEEK_END);
+		size = ftell(file);
+		fseek(file, 0, SEEK_SET);
+		azicfg = (unsigned char*)malloc(size);
+		fread(azicfg, size, 1, file);
+		fclose(file);
 
-	snd->configSyncAudio = azicfg[0] ? true : false;
-	snd->configForceSync = azicfg[1] ? true : false;
-	snd->configAIEmulation = azicfg[2] ? true : false;
-	snd->configVolume = azicfg[3];
-	free(azicfg);
-
+		snd->configSyncAudio = azicfg[0] ? true : false;
+		snd->configForceSync = azicfg[1] ? true : false;
+		snd->configAIEmulation = azicfg[2] ? true : false;
+		snd->configVolume = azicfg[3];
+		free(azicfg);
+	}
+	
 	snd->AI_Startup();
 
 	return TRUE;
@@ -354,12 +357,14 @@ INT_PTR CALLBACK ConfigProc(
 
 			FILE *file;
 			file = fopen("Config/AziCfg.bin", "wb");
-			fprintf(file, "%c", snd->configSyncAudio);
-			fprintf(file, "%c", snd->configForceSync);
-			fprintf(file, "%c", snd->configAIEmulation);
-			fprintf(file, "%c", snd->configVolume);
-			fclose(file);
-
+			if (file != NULL)
+			{
+				fprintf(file, "%c", snd->configSyncAudio);
+				fprintf(file, "%c", snd->configForceSync);
+				fprintf(file, "%c", snd->configAIEmulation);
+				fprintf(file, "%c", snd->configVolume);
+				fclose(file);
+			}
 			break;
 		case IDCANCEL:
 			EndDialog(hDlg, 0);
--- a/AziAudio/main.cpp
+++ b/AziAudio/main.cpp
@@ -156,21 +156,41 @@ EXPORT Boolean CALL InitiateAudio(AUDIO_INFO Audio_Info) {
 //	if ( (DirectSoundEnumerate(DSEnumProc, NULL)) != DS_OK ) { printf("Unable to enumerate DirectSound devices\n"); }
 
 	// TODO: Move from SoundDriver to a configuration class
-	snd->configAIEmulation = true;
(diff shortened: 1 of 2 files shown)


AziAudio/main.cpp
```diff
@@ -156,22 +156,42 @@ EXPORT Boolean CALL InitiateAudio(AUDIO_INFO Audio_Info) {
...
+	FILE *file;
+	file = fopen("Config/AziCfg.bin", "ab+");
+	fseek(file, 0, SEEK_END);
+	size = ftell(file);
+	if (size == 0)
+	{
+		fprintf(file, "%c", 1); // snd->configSyncAudio
+		fprintf(file, "%c", 0); // snd->configForceSync
+		fprintf(file, "%c", 1); // snd->configAIEmulation
+		fprintf(file, "%c", 0); // snd->configVolume
```
cxd4: Technically you could just `fputc` since all `fprintf` does is fputc() strings character-by-character.

```c
fputc(0x01, file); // snd -> configSyncAudio
fputc(0x00, file); // snd -> configForceSync
fputc(0x01, file); // snd -> configAIEmulation
fputc(0x00, file); // snd -> configVolume
```

This may not really be any better since compilers are usually smart enough to compile something like `printf("Hello, world!\n");` into actually `puts("Hello, world!");` for you, since saying "printf" instead sometimes yields preferred syntax and familiarity while the compiler optimizes it away from actually executing a full-blown printf() call (with all the extra steps involved in character conversions within the routine).

Also, less important, but if those settings are all just true/false, you theoretically could have put all four of them in a single byte instead of storing them as four single bytes.

Frank-74:
I check if file size is 0 (doesn't exist), to write file with default settings.

I'll add standard, more robust error handling later today.

Azimer:
I see that but you are using ftell and fseek which won't work if fopen failed.  fopen can fail for a lot of different reasons when it goes out in the wild.  Someone decides it's brilliant to put PJ64 on readonly media or a write protected folder network folder.  Or it's run from another emulator without a Config folder.  I am sure there are even more scenarios.  The key with software development is to prevent runtime exceptions from becoming unhandled runtime errors by enforcing your assumptions.  I appreciate your efforts Frank!
+	fseek(file, 0, SEEK_END);
+	size = ftell(file);

Frank-74:
Something like this?

    file = fopen("Config/AziCfg.bin", "ab+");
    if(!file)
    {
        printf("Error: Can't open file);
        exit(-1);
    }
Off the top of my head, not tested, not sure about exit condition, or printing error.  I need to read some docs to refresh my memory.  Been a while since I did C file operations.

Azimer:
Thinking something like:
`	int size;
	unsigned char *azicfg;
	FILE *file;
	file = fopen("Config/AziCfg.bin", "ab+");
	if (file != NULL)
	{
		fseek(file, 0, SEEK_END);
		size = ftell(file);
		if (size == 0)
		{
			fprintf(file, "%c", 1); // snd->configSyncAudio
			fprintf(file, "%c", 0); // snd->configForceSync
			fprintf(file, "%c", 1); // snd->configAIEmulation
			fprintf(file, "%c", 0); // snd->configVolume
		}
		fseek(file, 0, SEEK_SET);
		azicfg = (unsigned char*)malloc(size);
		fread(azicfg, size, 1, file);
		fclose(file);

		snd->configSyncAudio = azicfg[0] ? true : false;
		snd->configForceSync = azicfg[1] ? true : false;
		snd->configAIEmulation = azicfg[2] ? true : false;
		snd->configVolume = azicfg[3];
		free(azicfg);
	}
	else
	{
		snd->configSyncAudio = true;
		snd->configForceSync = false;
		snd->configAIEmulation = true;
		snd->configVolume = 0;
	}`

and

`			FILE *file;
			file = fopen("Config/AziCfg.bin", "wb");
			if (file != NULL)
			{
				fprintf(file, "%c", snd->configSyncAudio);
				fprintf(file, "%c", snd->configForceSync);
				fprintf(file, "%c", snd->configAIEmulation);
				fprintf(file, "%c", snd->configVolume);
				fclose(file);
			}`

Frank-74:
There's a problem with that.  It's now not reading the settings, on opening config, they're wrong/random.
They're not being set here:
		snd->configSyncAudio = azicfg[0] ? true : false;
		snd->configForceSync = azicfg[1] ? true : false;
		snd->configAIEmulation = azicfg[2] ? true : false;
		snd->configVolume = azicfg[3];
The file gets written correctly.  I need to work out a better solution.

cxd4:
What's up with this GitHub "review" comment system?  I can't reply to anything or quote anyone's code without starting a "review" and making a big production over something I just feel like nitpicking over.

Yes, the return value of calling fopen() should always be taken advantage of in secure code.

Same is true with fseek(), which unlike rewind() returns a value for error-checking.  If the standard C library function returns a value, you should take advantage of it to the fullest.

>Off the top of my head, not tested, not sure about exit condition, or printing error.

Is that for `(file == NULL)` after trying to open file for write-access?  If so, there's pretty much nothing you can do, other than maybe warn the user to run the emulator off something that's not write-protected or check the file wasn't marked as "Read Only".

You could fall back to writing it somewhere else, like AppData or other OS-specific folders, but this adds to the recursion of where you have to check when fopen'ing for READ access, doubling what you have to implement.  I'd just tl;dr it and leave it use the current directory.

I don't know about that `exit(-1)` you had, though.  I wouldn't raise a program abortion termination status over merely failing to save changes to the config file.

>I need to read some docs to refresh my memory. Been a while since I did C file operations.

Nah I always forget them myself too.  I always look up the docs every single time I need to use them. :)

Azimer wrote:

>I see that but you are using ftell and fseek which won't work if fopen failed.

Oh that's another potential issue, with portability.

The C standard does not regulate that fseek() to the end of the stream and ftell() will yield the file size in bytes.  POSIX does, yes, but not all C platforms are POSIX-compliant.

Getting the file size in bytes from a currently opened stream is a little bit more tedious to do in a universally portable way.  If you don't need to know it for anything, don't implement it.

cxd4:
```c
	int size;
...
	size = ftell(file);
```
Also, as a matter of principle, size-purpose-like variables should try to be declared as the `size_t` type, not `int`.  `int` is more of a general-purpose arithmetic/logic type, but if this "size" variable is to be used as an array index, memory allocation tests or boundary indexing, it is the kind of size that should be a `size_t`.

The unfortunate exception is that ftell() actually returns `long`, so in this case that would be best.  It's still larger than `int` and is capable of handling larger file stream positions so is more secure.

Frank-74:
Ok, I've sorted it out.  Now it only writes the file when you change settings and click ok.  It just reads the file if it exists on open and applies the settings.  Updating PR now.

Frank-74:
Did you also want a check on the configVolume setting?  To make sure it's less than 100?

cxd4:
Volume probably should be clamped.

```c
if (volume < -100)
    volume = -100;
if (volume > +100)
    volume = +100;
```
...or does it start at 0; I don't know.

If you feel like having fun with a text-based configuration format too I'd be happy to support and/or help with that.  I know I used binary-based bit-wise config for my RSP plugin...that was really more of a thing with not wanting to link in msvcrt.dll on the Windows build.  If you prefer binary though that's cool too.

Frank-74:
@cxd4 
Volume is 0 for max, 100 for mute.  It defaults to max, dialog control clamps it between 0-100 anyway.  So it's probably not needed.

Azimer:
I think validating the volume on read would be a good idea in practice.  If you don't make the change it will be made in a later PR.  @cxd4 If my memory is correct the slider is from 0 to 100.  In actual implementation, 0 is full volume and 100 is minimum.

Frank-74:
I'm done with changes for now.  Need to cook some dinner.  Cheers.

budtz:
I still cant seems to get saving audio volume to work. shouldnt it be fixed?

Frank-74:
@budtz
Yes it should.  Make sure to click OK in the config dialog to save the setting.  Does my DS8 build work?
https://www.dropbox.com/s/s4f9eju891l0zxt/AziAudioDS8.dll?dl=1

budtz:
that works

cxd4:
>It was shorter and easier to edit with binary, only 1 byte per setting.

True, but it would be further still shorter to have 1 bit per setting (except for non-Boolean things like the volume of course).

The hard thing about text-based is you have to invent your own line-parsing functions to read left and right of things like an `=` character, but once that's done your code becomes self-explanatory by searching for strings that self-document what the purpose of that config is (unlike bits and bytes in binary-based storage where you have to //comment what each field for).

But like I said binary is good too; it was just an alternative I was thinking of.  I probably won't switch to text-based config for my RSP either due to a) the lack of conceivable things to configure b) needing msvcrt.dll on Windows.

PR #159, opened 2017-02-08
secure, simpler config file initialization

cxd4:
A number of things being addressed:

1.  Calling `fread(azicfg, size, 1, file)` without a return value is bad.  That means if the config file was only 3, 2 or less than `size` bytes, no bytes would be imported at all.  Switching it around to `fread(azicfg, 1, size, file)` would have at least written what was available, but you should still check for the return value to see how many bytes were accessed without errors.
2.  Using `malloc` based on the file size of the config file could mistakenly allocate several GB of RAM if the config file name happened to represent some binary file other than the intended AziAudio configuration.  If Config/AziCfg.bin was really some other file just named wrongly and was say 4 GB in size, we'd really only need to malloc() just the first 4 bytes.
3.  Trying to calculate the file size with fseek(...SEEK_END) is undefined and only correctly seeks to the end of the file in byte units on POSIX-conformant implementations.  A portable implementation of getting the file size is more difficult, tedious and unneeded for the task of just initializing the config.
4.  We don't need fread(), at least not yet.  As in the config file we're only working with the plain and simple UCHAR type, we can just loop through the file safely with fgetc() and secure error checks for each byte read.

Now the volume clamp I don't know about if I should have bothered doing.  I understand it's 0 max 100 min for some sort of Windows sound API (maybe both DirectSound and XAudio), but if the user wanted more precise audio control in some other API like OpenAL or SDL a `float` or `double` might have been neater, then it could be parsed to DirectSound-specific volume scales within the DirectSound code converting it to an integer between 0 and 100.  But YOLO, did it anyway.

### DIFF
--- a/AziAudio/main.cpp
+++ b/AziAudio/main.cpp
@@ -168,33 +168,34 @@ EXPORT Boolean CALL InitiateAudio(AUDIO_INFO Audio_Info) {
 	DMEM = Audio_Info.DMEM;
 	IMEM = Audio_Info.IMEM;
 
-	int size;
-	unsigned char *azicfg;
+	size_t file_size;
+	int character;
+	unsigned char azicfg[4];
 	FILE *file;
 	file = fopen("Config/AziCfg.bin", "rb");
 	if (file == NULL)
 	{
-		snd->configSyncAudio = true;
-		snd->configForceSync = false;
-		snd->configAIEmulation = true;
-		snd->configVolume = 0;
+		azicfg[0] = TRUE;
+		azicfg[1] = FALSE;
+		azicfg[2] = TRUE;
+		azicfg[3] = 0; /* 0:  max volume; 100:  min volume */
 	}
 	else
 	{
-		fseek(file, 0, SEEK_END);
-		size = ftell(file);
-		fseek(file, 0, SEEK_SET);
-		azicfg = (unsigned char*)malloc(size);
-		fread(azicfg, size, 1, file);
-		fclose(file);
-
-		snd->configSyncAudio = azicfg[0] ? true : false;
-		snd->configForceSync = azicfg[1] ? true : false;
-		snd->configAIEmulation = azicfg[2] ? true : false;
-		snd->configVolume = azicfg[3];
-		free(azicfg);
+		for (file_size = 0; file_size < sizeof(azicfg); file_size++) {
+			const int character = fgetc(file);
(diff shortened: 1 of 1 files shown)


Frank-74:
Hahaha, you've gone postal on the config.  Nice one...:)

You forgot per game settings!

Only joking...

cxd4:
I actually wouldn't mind implementing that.

It would probably demand switching to text-based config format, though.  Per-game configuration settings is possible with binary file format only in a vague sense.  It would counter-intuitively go against the beneficial fixed nature of binary encoding.

Probably doesn't matter.  Needing settings to be on and off for different games is sort of evidence of hacking around things anyway.

Azimer:
I am not a fan of game specific hacks.  I try for a one size fits all approach.

Frank-74:
Would it be possible to reserve some bytes in the dll to save the config settings?  So we don't need to write a file.  Could probably get away with just 1 byte, and use bits.

cxd4:
I think a separate file would be best.  I'm not even sure if it's possible on Windows for a DLL to modify itself, granted that the DLL is already running.  That would be self-modifying code and Windows would need to implement Protect Memory with ABL to support that (unless you just want to install Linux to have it natively).

A DLL being able to save settings into itself would solve the issue of having to worry about where to create a new file...but the DLL itself could be write-protected or on a CD-ROM so there would be no invincibility there either.

Frank-74:
int character; is not needed.
1>main.cpp(171): warning C4101: 'character': unreferenced local variable

cxd4:
yea it's not...I spotted that after I came home from work.

The reason is it's declared twice.
https://github.com/Azimer/AziAudio/blob/7036e7234b6f1dcddf5e5f7be53a0e1bfcfad027/AziAudio/main.cpp#L172
https://github.com/Azimer/AziAudio/blob/7036e7234b6f1dcddf5e5f7be53a0e1bfcfad027/AziAudio/main.cpp#L186

So it defaults to the local declaration and ignores the parent-scope one.  I just hadn't bothered with a PR because it seemed trivial, but I guess I should.

Removing either one is fine.

Frank-74:
I built from the NewAudio branch, and the configure audio dialog never opened.  Master is fine.

Azimer:
I will delete my branch and recreate once I am completed with testing to avoid such issues and potential merge conflicts.

PR #168, opened 2017-02-10
warning: logical NOT applied to left side only [-Wlogical-not-parentheses]

cxd4:
```
./../AziAudio/SoundDriver.cpp: In member function 'void SoundDriver::AI_ResetAudio()':
./../AziAudio/SoundDriver.cpp:170:40: warning: logical not is only applied to the left hand side of comparison [-Wlogical-not-parentheses]
  m_audioIsInitialized = (!Initialize() == TRUE);
                                        ^
```

### DIFF
--- a/AziAudio/SoundDriver.cpp
+++ b/AziAudio/SoundDriver.cpp
@@ -119,7 +119,7 @@ void SoundDriver::AI_Startup()
 #ifdef LEGACY_SOUND_DRIVER	
 	if (m_audioIsInitialized == true) DeInitialize();
 	m_audioIsInitialized = false;
-	m_audioIsInitialized = (Initialize() != FALSE) ? true : false;
+	m_audioIsInitialized = (Initialize() != FALSE);
 	if (m_audioIsInitialized == true) SetVolume(configVolume);
 #else
 	Initialize();
--- a/AziAudio/SoundDriver.cpp
+++ b/AziAudio/SoundDriver.cpp
@@ -119,7 +119,7 @@ void SoundDriver::AI_Startup()
 #ifdef LEGACY_SOUND_DRIVER	
 	if (m_audioIsInitialized == true) DeInitialize();
 	m_audioIsInitialized = false;
-	m_audioIsInitialized = (!Initialize() == TRUE);
+	m_audioIsInitialized = (Initialize() != FALSE) ? true : false;
 	if (m_audioIsInitialized == true) SetVolume(configVolume);
 #else
 	Initialize();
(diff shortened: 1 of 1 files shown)


AziAudio/SoundDriver.cpp
```diff
@@ -119,7 +119,7 @@ void SoundDriver::AI_Startup()
 #ifdef LEGACY_SOUND_DRIVER	
 	if (m_audioIsInitialized == true) DeInitialize();
 	m_audioIsInitialized = false;
-	m_audioIsInitialized = (!Initialize() == TRUE);
+	m_audioIsInitialized = (Initialize() != FALSE) ? true : false;
```
Azimer: I understand (!Initialize() == TRUE) has a warning.  But I disagree with the solution being elegant.  Initialize() appears to return a BOOL of TRUE or FALSE.  So wouldn't the best solution be 
```C
m_audioIsInitialized = (Initialize() != FALSE);
```
It will by default return a true boolean of true or false.  Is this considered bad from a -Wall of crazy PoV?  This won't stop the merge but I respect your opinion and a solid debate.

cxd4: That is a good question, thanks.  I will check in a minute.

I'm told that for compatibility reasons something like x = (y < z)...C90 and C99 natively evaluate (y < z) to 1 or 0, not C++'s `true` or `false`.  I did a ternary to `(y < z) ? true : false` just to be explicit, but will it be handled either way?  Could be...if so I can push another commit to this.  I'll check.

cxd4: I just pushed a second commit.

It looks like it doesn't matter.  Whether I say:
```c
m_audioIsInitialized = (Initialize() != FALSE)
m_audioIsInitialized = (Initialize() != FALSE) ? true : false
m_audioIsInitialized = (Initialize() != FALSE) ? 1 : 0
```

... `-Wall` is not strict enough to nitpick any warnings about any of those three forms.

I still think ? true : false is more guaranteed to be accurate, but the 1/0 logic only applies to C and this is a C++ source file anyway.  It's not worth picking over.

cxd4:
This would seem to be a bug fix to the legacy sound driver code.

The original:
```c
(!Initialize() == TRUE);
```
... was being parsed as (Initialize() == FALSE) == TRUE, or just (Initialize() == FALSE).

cxd4:
Wait, wtf?

Don't merge this yet.  XAudio and DSound Initialize() examples show that FALSE should be returned on success, not failure.  So this was backwards of backwards; I need to invert that.

cxd4:
There, that's done.

But at this point, I think somebody might see the `m_audioIsInitialized = (Initialize() == FALSE);` and think that's a serious typo.  Most people would assume a function returns true on success and false on errors.  Are ya sure a ternary form of `m_audioIsInitialized = (Initialize() == FALSE) ? true : false;` wouldn't be somewhat definitive and readable? :P Nah, it's not that important.

PR #174, opened 2017-02-10
Clean up dialog mute logic

Frank-74:
Mute setting is not needed in config.  Just check the volume, set the checkbox if volume=100, and check the checkbox setting, set the volume to 100 if checked.


cxd4:
>_No description provided._

That is some seriously clean mute dialog logic.

Oh, wait, the code's on the Files Changed tab.  (Never mind, bad 2 AM jokes. :P)

Frank-74:
7:07 am here !  Been up all night, sleep time....

Azimer:
Does this actually work without issuing a snd->SetVolume?  I am going to have to pull your code base and test before I merge.  I am skeptical it will retain its functionality.

Azimer:
I just tested.  This breaks the intended functionality.  You should be able to go into the configuration dialog, set your volume slider and have it affect your sound volume while it tracks.  Also, when you mute then unmute, the volume should be reset to the previous selection rather than full volume.  The functionality presented here is not what I intended.  If you can find another way to keep it clean and retain the functionality, I will continue with the merge.

Frank-74:
Ok, it now sets the volume immediately.

Azimer:
I see you closed it.  Do you intend to reopen it?

Azimer:
I think for mute, you just need to disable the slider, set the volume for the device to mute.  Then for unmute, enable the slider and set the volume for the device depending on the slider.  This still achieves what you are intending.

PR #175, opened 2017-02-11
Remove configMute setting

Frank-74:
Redone, as last PR was a mess.  Volume is set immediately with any changes to slider.  Mute checkbox mutes volume and restores previous set volume when checked/unchecked.


Azimer:
Ignore my other comments.  I will take a look at this.

Frank-74:
Quick question.  Are the HLE and RSP settings needed?  The configHLE setting is always true it seems.  Was wondering about the if (snd->configHLE) HLEStart (); condition when using LLE setting in emulator.

I was wanting to redo the dialog and add a TAB control.  With sync and volume on tab1, device, backend selection and audio logging on tab2, shortcuts on tab3

Azimer:
No it's not needed since that is set by PJ and there's no other active PJ spec emulator.

PR #176, opened 2017-02-11
Remove configHLE and configRSP

Frank-74:
Removes redundant HLE and RSP settings.


Azimer:
I'll merge this.  I still need to look at the other one.  The intention of disabling HLE was when having a RSP compiler internally was still in the realm of possibilities.  I decided since then to limit the design of the project to focus just on audio playback keeping the HLE around more for legacy.

Issue #178, opened 2017-02-12
HLE volume problem


Frank-74:
HLE volume mix is wrong.  Affects lots of games.

Here is example of Road Rash 64, HLE and LLE to compare.
https://www.dropbox.com/s/izqhfcxq92tkpok/RoadRashHLE.wma?dl=1
https://www.dropbox.com/s/shhq3h6lc67bs33/RoadRashLLE.wma?dl=1

Rubbish sound recorder in Windows 8.1, only records at 96Kbps bitrate, so quality is bad, but you get the idea.

LegendOfDragoon:
By any chance were you loading a savestate?

Frank-74:
Nope.

Edit game settings, selected XA2 plugin, checked HLE audio, start game, record audio, end emulation.
Edit game settings, unchecked HLE audio, start game, record audio, end emulation.

Frank-74:
Other games with very noticeable HLE dry/wet mix problems.
BattleTanx : The "3do" speech is very quite compared to following music at startup.
BattleTanx - Global Assault : Music has echo effect which overpowers dry signal, same as Road Rash.
Castlevania - Legacy of Darkness : Breathing sounds missing, and lots of sounds missing, volumes totally wrong.  Example of the start intro here:
https://www.dropbox.com/s/gtmioxayqgsqyu3/CastlevaniaLOD-HLE.wma?dl=1
https://www.dropbox.com/s/ri32i1go8vjfbls/CastlevaniaLOD-LLE.wma?dl=1

cxd4:
>It would be very useful to get audio logging to work in the plugin.

I would be happy to implement that once I am able to run the plugin.

Also, for some reason those Dropbox links don't let me play the audio in-line through the browser.  I have to download both of them as files to play them locally.  th;dd

Frank-74:
>Also, for some reason those Dropbox links don't let me play the audio in-line through the browser. I have to download both of them as files to play them locally. th;dd

I've just downloaded and installed Audacity, re-recorded, and saved as mp3, they now play in browser.
https://www.dropbox.com/s/3s4ay28abjwyyy9/CastlevaniaLOD-HLE.mp3?dl=0
https://www.dropbox.com/s/5j3l5lb9wdm476p/CastlevaniaLOD-LLE.mp3?dl=0

vgturtle127:
@Frank-74 Even LLE doesn't sound right to me, though. 😢

Frank-74:
[...]

PR #180, opened 2017-02-13
Fix HLE volume problem #178

Frank-74:
Details of bug here: #178 .


cxd4:
That was what I was supposed to do.

It's tricky because we're emulating a big-endian CPU on a little-endian CPU.  If we were compiling AziAudio for the N64 machine, *(s16 *)&k0 would be grabbing the upper 16 bits.  Since we're on x86, it's grabbing the upper 16 bits anyway but using memory assumptions.

PR #189, opened 2017-03-03
moved UNREFERENCED_PARAMETER workaround ifndef WIN32

cxd4:
Something about recent branches merged into master has removed the includes for these defines.


Azimer:
Thanks for all your changes.  Sorry for the delay -- I am 800 miles away from home helping my mother-in-law move.

cxd4:
Gotcha.  Having moved a lot I know that can be a time-consuming thing.

Issue #184, opened 2017-03-02
Very slow with PJ64 "Sync using audio" option.


Frank-74:
Tested with latest master.  Turning on Sync using audio in Project64, causes VI's to drop dramatically (15 in Goldeneye), and eventually PJ64 stops responding and I have to close it with task manager.

Azimer:
I have no idea how that is implemented.  I imagine turning on sync in Project64 as well as having either of the two options set in AziAudio will cause a conflict.  Trying turning off Prevent Buffer Overruns and Force Old Audio Sync.  Those are the only two options that stall the emulator to help bleed off some of the audio buffer.

Frank-74:
I have all the sync options turned off in AziAudio.

This never happens building from New Audiocode 181 commit.  https://github.com/Azimer/AziAudio/commit/efc7c73ba4813f2b9f9ca6cf67b3d89502ab933a .

Frank-74:
Removing this Sleep(1) from XAudio2SoundDriver.cpp fixes this issue.

https://github.com/Azimer/AziAudio/blob/master/AziAudio/XAudio2SoundDriver.cpp#L237

It also fixes VI's lagging and sound getting ahead.  Very noticeable in Body Harvest.
And fixes problem I had with musyx games needing a bigger buffer.

theboy181:
release?

Frank-74:
Here's my build with this Sleep(1) only turned on when Force Audio Sync is enabled, instead of removing it totally.  I also changed Backend and Buffer FPS to 95, slightly smaller buffer, but it's cleaner with Body Harvest.

Make sure Prevent Buffer Overruns is on, and Emulate Audio Interface is on, Force Audio Sync OFF.
FAT off for Body Harvest.

https://www.dropbox.com/s/9wv7nohyiz24mtf/AziAudioXA2.dll?dl=1

Frank-74:
@theboy181
Try TWINE with FAT off and disable PJ64's framelimiter for good example of the audio staying clean at mega fast speed, and in time.

theboy181:
Sounds pretty good.. Ill leave the plugin as default and report any issue I find. Do those settings need to be set every time? Is it the best settings for most games?

Frank-74:
Leave Prevent Buffer Overruns and Emulate Audio Interface ON for every game.  Plugin remembers your settings.

[...]

PR #191, opened 2017-03-24
[XA2] Fix stutter/stalling/crackling with smaller buffer?

Frank-74:
Sleep(1) here causes emulation to stall.  I removed it and I no longer need a bigger buffer for musyx games.  I changed bufferFPS to:
	configBufferLevel = 2;
	configBufferFPS = 96;
	configBackendFPS = 96;
Which seems slightly better than 90 here.
Also with Sleep(1) here, Body Harvest stalled the VI's to 40+ in intro (in places), and audio got ahead of gfx badly (with latest WIP GLideN64).  Body Harvest still crackles bad with FAT, but is now perfect with FAT off.

TWINE, RE2, WDC and Top Gear Rally are all good with FAT on or off.  PBO and AI on in AziAudio.  Though some games PBO don't sync to 60 with framelimiter off, Twisted Edge, WDC, Top Gear Rally, Banjo Kazooie + Tooie, Goldeneye, Perfect Dark, Conker's Bad Fur Day, Winback, BattleTanx, Diddy Kong Racing and Donkey Kong 64 are a few games I noticed.  Most other games do sync to 60 with framelimiter off and PBO on.

### DIFF
--- a/AziAudio/XAudio2SoundDriver.cpp
+++ b/AziAudio/XAudio2SoundDriver.cpp
@@ -234,7 +234,6 @@ DWORD WINAPI XAudio2SoundDriver::AudioThreadProc(LPVOID lpParameter)
 				g_source->GetState(&xvs);
 			}
 		}
-		Sleep(1);
 	}
 	return 0;
 }


Frank-74:
This also fixes #184.

Frank-74:
Some games don't sync with smaller buffer because of this condition in SoundDriver.cpp

```cpp
if (length < m_MaxBufferSize && Configuration::getSyncAudio() == true)
```

I've removed the length < m_MaxBufferSize and it syncs with all games fine with framelimiter off.

Made a new tree here to compare:
https://github.com/Frank-74/AziAudio/tree/Fix-Sync-Audio

Azimer:
I added the length < MaxBufferSize to prevent WDC and TGR from skipping frames.  It's really unnecessary and causes too many problems...  hence the exception.

As far as removing Sleep(1) - I think Sleep(0) is a better idea.  Even better may be SleepEx(1) and allow the thread to accept messages to wake up when the play buffer ends.  Thoughts?  I certainly don't expect you to perform that work... was just curious of your thoughts on the matter.

I am not too keen on messing with the FPS change too much.  It seems awkward... but I am willing to make them into "Advanced" configuration settings.

Sorry I've been away so much.  I haven't had a lot of personal time the last two weeks.  I've been gone the last two weekends and my wife is having some medical issues.  Let me know your thoughts on the above or if you want to chat about them, I plan on being on freenode #Project64 the rest of the evening (next 4 hours or so) and most of tomorrow.

Frank-74:
FPS at 95 was the only setting where Body Harvest didn't have any crackles with FAT off.  I couldn't get Body Harvest to not crackle with FAT on, tried FPS from 30 to 100.

I've messed with the options a bit for testing.  Added a double size buffer option and audio thread sleep option  (Sleep(0) / Sleep(1)).  It's easier to see/hear the differences without having to keep rebuilding.

Can give SleepEx(1) a try.  How does it work, (milisecond, true/false)?  I'm not sure how to make it work, setup a bool variable and make it true when buffer ends?  I don't know for sure, I'll leave that to an expert.

Cheers.

PR #192, opened 2017-03-25
Update Xbox port

weinerschnitzel:
Please review the commit Create SoundDriverFactoryXBox.cpp for discussion on how we should best handle creating new SoundDriverFactory.cpp's and where you think they should belong. This could be an issue for Linux, as well. The 3rd party folder could be restructured to be a Platform folder, but that would involve changes to references for the directx folder.

### DIFF
--- a/AziAudio/main.cpp
+++ b/AziAudio/main.cpp
@@ -27,7 +27,6 @@
 	#include <ios>
 #endif
 
-using namespace std;
 
 SoundDriverInterface *snd = NULL;
 
@@ -101,7 +100,7 @@ EXPORT Boolean CALL InitiateAudio(AUDIO_INFO Audio_Info) {
 		delete snd;
 	}
 
-	snd = SoundDriverFactory::CreateSoundDriver(SoundDriverType::SND_DRIVER_XA2L);
+	snd = SoundDriverFactory::CreateSoundDriver(SND_DRIVER_DS8);
 
 #ifdef USE_PRINTF
 	RedirectIOToConsole();
--- a/AziAudio/AziAudio.xbx.vcproj
+++ b/AziAudio/AziAudio.xbx.vcproj
@@ -190,22 +190,31 @@
 			Filter="cpp;c;cxx;def;odl;idl;hpj;bat;asm;asmx"
 			UniqueIdentifier="{4FC737F1-C7A5-4376-A066-2A32D752A2FF}">
 			<File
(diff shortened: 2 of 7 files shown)


Azimer:
Hi!

In general I think I'd rather not have an XBox or Linux specific Factory but rather have them ifndef'd out.

AziAudio/main.cpp
```diff
@@ -27,7 +27,6 @@
 	#include <ios>
 #endif
 
-using namespace std;
```
Azimer:
```C
#ifdef USE_PRINTF
using namespace std;
#endif
```

AziAudio/main.cpp
```diff
@@ -101,7 +100,7 @@ EXPORT Boolean CALL InitiateAudio(AUDIO_INFO Audio_Info) {
 		delete snd;
 	}
 
-	snd = SoundDriverFactory::CreateSoundDriver(SoundDriverType::SND_DRIVER_XA2L);
+	snd = SoundDriverFactory::CreateSoundDriver(SND_DRIVER_DS8);
```
Azimer: This works?  If so, that's broken.  It should need to be qualified unless that's something weird in the version of studio you need to use for XBox.  What version is that?  I was pretty sure it had to be fully qualified to access the enum.

Azimer:
Revert bc44a2778b4a28a04ccaa3b1233e28482c495e8f and 7aef27acc7135d64982b23c4616dcbe07bc25c67.  Then rebase your changes with master.  You should have no issues then.

Azimer:
Lastly...  DS8 is an XBox default...  we need to find the best way to default that for your system.  Do you have a way for the user to change configuration or do you just use the default stuff?  This might be a good place for an ifdef to force the use of DS8 for XBox only.  I could look into this also if you'd like.  I haven't completed the configuration aspect yet.

weinerschnitzel:
Sounds good. I should have noticed the USE_PRINTF issue. The enum issue is with VS2003, I can't exactly explain why it is not valid as it was. The compiler is very old so I would like to blame it on a limitation.

I don't have a way to edit audio plugin settings within UI. I try to have those configured by editing files so the user can't mess with them. There are a lot of settings that are best set once, so I try to make it simple. If you were to have the configuration file assign a driver, this should work well for all parties.

What about adding SDL to SoundFactory and in windows builds in general? Perhaps still with an ifdef so SDL will not have to be included with the build?

PR #193, opened 2017-03-26
Fix default config settings.

Frank-74:
Defaults weren't saved if AziCfg.bin didn't exist.


Azimer:
Thanks Frank.  This must have been an oversight!

PR #198, opened 2017-08-08
Emulate Windows types in Configuration.h.

cxd4:
This fixes quite a lot of compile errors on Linux--in too many dimensions to paste the list of which ones.

The `my_types.h` I'd written into the repository is extensive but does not concern itself with emulating types from `<windows.h>` for non-Windows platforms.  Come to think of it, there is no such header that does this really, as the AudioSpec.h (zilmar's audio spec header) was something I'd just rewritten to minimize/eliminate any need for Windows-specific types (which the plugin specs shouldn't need to depend on anyway in most cases).

With that being said I decided to lump the types in `Configuration.h` since that's the only file that seems to use Windows-specific types in the Linux port as far as I can see (besides `Configuration.cpp` which of course is just going to include `Configuration.h` anyway).


cxd4:
https://github.com/Azimer/AziAudio/pull/199

I was focused more on the huge jungle of errors GCC was flooding with and not on the fact that they all surrounded a single line of code that was just more WIN32 configuration stuff we don't need to port.

PR #202, opened 2017-10-08
Recreate PR #199.

cxd4:
Just to be on the safe side, I've re-uploaded the PR from https://github.com/Azimer/AziAudio/pull/199.

Not sure what happened.  Seems there may have been a merge conflict while Azimer was merging the others then he added a commit to my fork to fix the conflicts somehow.  It appears like it's still good, but to be safe I did a clean re-clone and re-fork of the repo to base these commits on.

### DIFF
--- a/AziAudio/Configuration.cpp
+++ b/AziAudio/Configuration.cpp
@@ -7,9 +7,9 @@
 
 #ifdef _WIN32
 #include <windows.h>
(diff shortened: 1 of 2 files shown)


Azimer:
I agree.  I think the change I made caused issues or it didn't merge right from the WebUI.

PR #205, opened 2017-10-11
Configuration Changes

Azimer:
Allowed configuration to disallow sleeps in DS8 and XA2

Partial Commit - Working through GUID issues

Applied Backend to the configuration dialog

Also added a default driver selection and changed the registration to
include a friendly name for the driver.

Fixed up the property sheets a bit

Hotswap backends are not functional yet.

Last of the Configuration changes

Configuration changes finalized

Bug Fix


Frank-74:
I recently changed the DS8 disallow sleep condition, if (write_pos == last_pos), to if (write_pos == 0).
It reduced CPU usage by about a third.

Azimer:
If that works for you then it's better than no sleep.  I will make a similar change.  I didn't see as drastic a cpu reduction but this setting isn't designed for my system.

PR #208, opened 2017-10-12
Pull Request for WIP9

Azimer:
I have enough changed for WIP9.  Will focus on WIP10 as individual branches.


cxd4:
Looks like I caused a merge conflict with one if my #ifdef _WIN32 commits being merged.

Once I'm able to test developing and using the plugin on Linux I'd like to add sound driver options for SDL and OpenAL (both of which actually manage low-level sound systems like ALSA or OSS).  They are both also available on Windows and Mac but not something I would link in to the official public release of AziAudio since most Windows users are guaranteed to not have those dependencies installed to their system.

Azimer:
1000 ms / 1 frames per second
So if it's 1 frame per second... the latency is 1000 ms.
If it's 60 frames per second... the latency is 1000/60  (16.667ms).

This sets the buffersize in terms of time rather than bytes or frames which really mean nothing to the end user.  At least with time, they know 16ms would mean a latency of 16ms * # of buffers (bufferlevel).  No matter, it's an advanced feature I don't suspect people will really understand even if explained.

cxd4:
I'm sure that PortAudio would suffice.  It has to be at least (most likely more) as stable and sophisticated as OpenAL, although both are third-party packages not standard to the Slackware package tree that I would need to install externally.  Not really a significant factor.

A nice thing about SDL is I believe every modern Linux distro natively includes it without the need for adding or resolving additional dependency trees.  OpenAL is just cool because it resembles OpenGL...pretty under-developed though and taken over by proprietary existence.

